PR #9, opened 2015-04-27
Airavata API support for paginated result retrieval of Project and Experiment data

scnakandala:
No changes to the existing API methods are done. Paginated versions of API methods have been added to some existing API methods to retrieve project and experiment data. I have also added new unit test class for AiravataServiceHandler to test the API changes.


smarru:
Hi Supun, I see that you have added the mysql and derby scripts to teh api server. I understand they are needed to run your test cases, but if there any other way you could think to do this? having the database scripts at duplicate places will eventually complicate on which one to change.

scnakandala:
Hi Suresh,

I updated the pull request

PR #16, opened 2015-06-11
Fix for the script: generate-thrift-file.sh 

hasinitg:
I got an error when running the generate-thrift-file.sh due to the file paths that do not match the new source structure. It is fixed herewith and it worked for me after this fix.


smarru:
Hi Hasini, I have fixed the thrift script and also have updating all the generated files to thrift 0.9.2. Can you please hold onto to this PR until I commit these changes this morning. It will be great if you can take a PR after the thrift changes.

hasinitg:
Hi Suresh, please discard this if you have fixed the issue. Sure, I will update my copy of the source after your changes and before making the next PR.

PR #18, opened 2015-07-01
Add new experiment status records instead of overwriting previous records

scizzorz:


smarru:
Supun, can you review this and see if this can be safely merged onto 0.15 branch?

smarru:
@scnakandala can we incorporate this feature in new registry design

scnakandala:
Yes. We can do that @smarru

PR #21, opened 2015-07-13
Add field compute resource

dogless:
Goes in tandem with airavata-php-gateway pull request (https://github.com/apache/airavata-php-gateway/pull/5). A boolean field named, active, is added to the computeResource thrift model. Consequential setter and getter methods, and sql script modifications, are also added.


dogless:
I believe Supun actually already addresses this feature in master.

PR #23, opened 2015-07-19
Pull request containing the work carried out in the week of 12th-18th July

hasinitg:
Hi Suresh,

This pull request contains the following major changes:
1) enforcing security through thrift method interception supported by Guice AOP framework (as opposed to the in-method security enforcement) - I will document the details about this.
2) changing all API methods to accept AuthzToken (in both thrift definition file and AiravataServerHandler)
Please note that when changing the API method implementation in AiravataServerHandler,
- deprecated methods were not changed
- new method with "Internal" suffix was introduced for couple of methods which were called by other methods with in the API itself, in order to avoid multiple security checks on the same API call

3) Fixing the sample clients for compilation errors.

Thanks,
Hasini.


hasinitg:
Hi Suresh,

This is a gentle follow up on this PR. Since this pull request has code changes in number of places, it could soon become unmergeable with the other commits made to the master.

Thanks,
Hasini.

smarru:
Sorry for the delay hasini. This is now merged. I have some comments on guice, will continue the dev list conversation.

hasinitg:
Thank you Suresh.

PR #25, opened 2015-08-01
Added identity context and XACML based authorization for API calls

hasinitg:
This pull request contains the following:
1. Identity Context - which persists AuthzToken in a thread local
2. airavata-default-xacml-policy.xml which defines role based access control for admin and non-admin API methods.
3. XACML PEP (Policy Enforcement Point) to enforce fine grained authorization on the API calls.
4. Updated secure-client sample to showcase XACML based authorization on API calls w.r.t the default XACML policy. 
Appreciate if this could be merged with the master.
Thank you.
Hasini.


hasinitg:
This pull request is updated and now contains following items, in addition to the ones mentioned above:
1. Updated the airavata-default-xacml-policy in order to cater the current authorization requirements in PGA, as discussed in the dev list.
2. Updated the secure client sample to showcase the capabilities of the updated authorization policy.
3. Added the policy administration client (PAP) into the secure solution of airavata server, which publishes and enables the authorization policy in the WSO2 IS via its API, during the startup of airavata server - if the security is enabled.

Thanks,
Hasini.

PR #29, opened 2016-03-22
Adding SGE

kuytje:
1. Configure your SGE resource type in gfac yaml config file
   (airavata/modules/configuration/server/src/main/resources/gfac-config.yaml.config).
   As you can see you need to provide an output parser and email parser for
   your SGE, Unless you can't use existing implementations you need to come
   up with your own implementations of these interfaces.
2. Don't modify thrift generated file, instead add your new
   ResourceJobManagerType to
   airavata/thrift-interface-descriptions/data-models/resource-catalog-models/compute_resource_model.thrift
   file and generate thrift classes.
3. Come up with SGE JobConfiguration class, and modify
   org.apache.airavata.gfac.impl.Factory#getJobManagerConfiguration method
   to provide it. You should have xslt template for your SGE script file.
4. add your XSLT file in
   airavata/distribution/src/main/assembly/bin-assembly.xml. Otherwise it
   will not be added to bin directory.


kuytje:
Added sge, 
I ignored the maximum number of nodes value because I couldn't find the option in SGE and used serial as parallel environment as used on ROCKS clusters. Other users should create a parallel environment named serial using <qconf -ap serial>.

smarru:
Eric, apologies for not following on this PR for long. Can we revisit this? Are you using this SGE interface?

yasithdev:
Closing this PR (inactive)

PR #38, opened 2016-06-12
Added Xbaya UI

linfestus:
Adding the non-functional UI of Xbaya.

To run-
Run HomeWindow.Java


smarru:
@linfestus thanks for your contribution. This is looking good, keep up with the progress. 

I noticed this after merging the PR, do not commit the target folder to git. Also, always do a git rebase before you submit a PR.

PR #43, opened 2016-08-19
AMQPWSTunnel

jeffkinnison:
First implementation of the bridge between AMQP and WebSockets for streaming remote application data.


jeffkinnison:
As far as I can tell, the issue is not related to my AMQPWSTunnel module. It isn't built by maven (the module is entirely in Python), and the Travis CI build failure seems to be related to retrieving something from WSO2. Can we go ahead and merge this to get the module into Airavata?

PR #60, opened 2016-10-20
Add serialVersionUID for previous version of class

machristie:
To keep Credential.java compatible with the previous version, I'm adding
the serialVersionUID from the previous version. This will allow
serialized instances of the old version to be deserialized with the new
version.


smarru:
Thanks @machristie for this fix, I merged it into develop now

PR #75, opened 2016-11-10
AIRAVATA-2207 Admin users can't see experiment summaries

machristie:
Adds the following API server methods
* getExperimentByAdmin
* cloneExperimentByAdmin

These methods allow an admin to see and clone experiments for which they don't have the READ sharing permission.


machristie:
@scnakandala can you review this when you get a chance?

PR #77, opened 2016-11-17
AIRAVATA-2231 : Bug in UserResourceProfile.save() method 

anujbhan:
* Resolved save method bug, it now accepts (userId, GatewayID).
* https://issues.apache.org/jira/browse/AIRAVATA-2231


machristie:
@anujbhan should UserResourceProfile.java have `@Id` annotation on the `gatewayID` field?  I'm not that familiar with JPA so I'm just asking, I'm not sure.

PR #79, opened 2016-11-29
AIRAVATA-2232 Add CredentialOwnerType to credential model

machristie:
Added date stripping from Thrift generated stub classes to
credential-store-stubs/pom.xml.


machristie:
@scnakandala can you please review this when you have a chance?  I'm still new to developing with Thrift, etc. so please check to make sure I have implemented this correctly.  Thanks.

PR #81, opened 2016-12-01
Airavata 2245 Bug Resolution

anujbhan:


anujbhan:
https://issues.apache.org/jira/browse/AIRAVATA-2245

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/app/catalog/model/UserResourceProfilePK.java
```diff
@@ -0,0 +1,66 @@
...
+    public UserResourceProfilePK(String userId, String gatewayID) {
+        this.userId = userId;
+        this.gatewayID = gatewayID;
+    }
+
+    public UserResourceProfilePK() {
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
machristie: Why is hashCode and equals implemented like this? Two UserResourceProfilePK instances with the same userId and gatewayID should be equal and should hash to the same value, right?

machristie:
Other than my comment about, this looks good. Thanks for the fast work Anuj!

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/app/catalog/model/UserResourceProfilePK.java
anujbhan: @machristie ,
yes you are right, Even I can't figure out why this is the case, all the composite Key classes in registry are implemented this way. 

@scnakandala ,
Can you give some insight ?

scnakandala: I agree with Marcus. We should do it in a similar way as in expcatalog https://github.com/anujbhan/airavata/blob/1bb80426d1acaa6d65aa49951fbb5fd1bc6a1703/modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/experiment/catalog/model/ExperimentErrorPK.java#L57

machristie: Since it is this way in all of the registry, can we create a separate issue for cleaning this up and merge this PR?

scnakandala: +1

PR #85, opened 2016-12-13
AIRAVATA-1792 Implement ability for admin to search for experiments using the created user as well

machristie:
Adds the three more experiment statistics filtering options to the API
* username
* application id
* hostname id


machristie:
@scnakandala Can you review and merge this when you get a chance? Thanks.

PR #86, opened 2016-12-13
AIRAVATA-2163 Setting type on CredentialSummary

machristie:
Setting 'type' since it is required on CredentialSummary.


machristie:
@shamrath @anujbhan can you please review this when you get a chance? Thanks.

PR #89, opened 2016-12-16
AIRAVATA-2282 notification sort: Prevent int overflow

machristie:
Instead of returning the difference as a long and casting to an int,
which risks integer overflow, just return 1 or -1 corresponding to the
sign of the difference.

Note: this branch was branched from master which is ahead of develop by a few spurious commits.  My change for this bug fix is really just NotificationRegistry.java change.


machristie:
@scnakandala  can you review this when you get a chance? Thanks.

PR #104, opened 2017-04-04
Temporary fix for mysql-selinux issue

ECoulter:
Tmp fix for this issue: https://jira.mariadb.org/browse/MDEV-12231

Adds selinux policy module to allow mysql to set uid/gid. Also
checks for .pp file which should be incuded in future releases
of MariaDB.


yasithdev:
Closing this PR (inactive)

PR #122, opened 2017-09-26
Initial Commit for Airavata PGA Dockerfile

hiteshkumardasika:
Created Dockerfile for Airavata PGA Installation
Dockerfile still needs lot of changes. Image can be built and container can be started but still there are some configurations that need to be changed


dev-tools/docker/Dockerfiles/pga_config.php.j2
```diff
@@ -11,88 +11,80 @@ return array(
...
          * Tenant admin's username
          */
-        'admin-username' => '{{ admin_username }}',
+        'admin-username' => 'admin',
 
         /**
          * Tenant admin's password
          */
-        'admin-password' => '{{ admin_password }}',
+        'admin-password' => 'admin',
```
Ameya05: I don't think this file with sensitive content should be checked into repository. Although the passwords seem to be for your local installation, it would be better if u revert it to the {{ variable }} version.

hiteshkumardasika: But these configurations cannot be parameterized as these files directly go into the docker container through Dockerfile. To test the installation, these values are not parameterized.

yasithdev:
Closing this PR as its superseded by commit #9e541b3.

PR #127, opened 2017-10-16
Added Docker files for [AIRAVATA-2093]

jeffreypeter:


dev-tools/docker/apiserver/airavata-server.properties
```diff
@@ -0,0 +1,358 @@
...
+###########################################################################
+
+#This will be the primary monitoring tool which runs in airavata, in future there will be multiple monitoring
+#mechanisms and one would be able to start a monitor
+monitors=org.apache.airavata.gfac.monitor.impl.pull.qstat.QstatMonitor,org.apache.airavata.gfac.monitor.impl.LocalJobMonitor
+
+#These properties will used to enable email base monitoring
+email.based.monitor.host=imap.gmail.com
+email.based.monitor.address=test@scigap.com
+email.based.monitor.password=changeme
```
Ameya05: Should this file be checked in ? It has a bunch of usernames and passwords in it.

jeffreypeter: This file was copied from
https://github.com/jeffreypeter/airavata/blob/develop/modules/configuration/server/src/main/resources/airavata-server.properties

So I think these test usernames and passwords.

PR #133, opened 2017-10-31
Created database related entities, repositories and exposed them using OpenJPA for thrift models to use. Added testc ases for the same too

harsha161992:


smarru:
@harsha161992 i think we can close this PR right? since you now have it in sandbox repo?

PR #134, opened 2017-10-31
Allocation manager notification system 

nikchett:


smarru:
this is now in sandbox

PR #141, opened 2017-11-14
[AIRAVATA-2591]: Add group roles capabilities to Groups

sachinkariyattin:
Currently, the group members do not have fine-grained roles so only the owner can add users to this group.
It will be better to add group-member-roles:
The **owner** who creates the group initially. This ownership should be transferable to a different user. Only owners should be allowed to delete the groups.
**admin** - a user role which will allow to add more users or remove users from the group.
**user** - all users added to the group are by default in this role.

* Create entities for each of the users (Thrift IDL)
* Generate thrift stubs
* Add handler methods (CPI's)

**Changes**
* Added the Admin and Owner models by making changes to the existing thrift files.
* Added the following CPI methods and generated the thrift stubs  
```
     transferGroupOwnership()  
     addGroupAdmins()  
     removeGroupAdmins()
     hasAdminAccess()  
     hasOwnerAccess()  
```
* Added new persistence entities - Admin and Owner


smarru:
@scnakandala can you please review this PR?

scnakandala:
@sachinkariyattin My understanding is that admins are for each domain. But not for each group. But from the implementation code I see that admin table will have a reference to a group. I think this should not be the case. Do you have a specific reason why you modelled the admin table like this?

smarru:
@scnakandala I am suggesting we have admins for each group. An example use case is airavata for class room. A domain level admin will create a group for class to be taught and will make a faculty who is teaching that class an admin. The faculty can further add AI/TA's as admins to the group. Only these admins can add other users to this group and so on.

scnakandala:
Makes sense @smarru. The pull request seems ok to me. I suggest that we change the names a little bit so that it won't create confusion. For eg. Admin -> GroupAdmin, Owner -> GroupOwner etc.

smarru:
@sachinkariyattin can you please change the names as supun suggested as propagate them all the way to the DB.  i can then merge this PR.

sachinkariyattin:
@smarru @scnakandala Thanks for the review. 
I will make the suggested changes and update the PR

sachinkariyattin:
@smarru I have made the changes

modules/sharing-registry/sharing-registry-server/src/main/resources/sharing-registry-mysql.sql
```diff
@@ -56,6 +56,21 @@ CREATE TABLE USER_GROUP (
   FOREIGN KEY (OWNER_ID, DOMAIN_ID) REFERENCES SHARING_USER(USER_ID, DOMAIN_ID) ON DELETE CASCADE ON UPDATE NO ACTION
 )ENGINE=InnoDB DEFAULT CHARACTER SET=latin1;
 
+CREATE TABLE GROUP_OWNER (
+  OWNER_ID VARCHAR(255) NOT NULL,
+  GROUP_ID VARCHAR(255) NOT NULL,
+  DOMAIN_ID VARCHAR(255) NOT NULL,
+  PRIMARY KEY (OWNER_ID, DOMAIN_ID),
```
machristie: This primary key restricts users from owning more than one group. Perhaps the primary key should be `(GROUP_ID, DOMAIN_ID)`, which would restrict groups from having more than one owner?

Related question: USER_GROUP has an OWNER_ID column.  Do we need a GROUP_OWNER table?

smarru: @sachinkariyattin i merged this PR before I reviewed @machristie comment above. After reviewing it back again, I fully agree with this above comments. Since groups can have multiple admins, it makes sense to have a GROUP_ADMIN table. But since groups can only have one owner, it makes sense to get rid of GROUP_OWNER table and use the OWNER_ID in the USER_GROUP table. Can you please make changes accordingly?

sachinkariyattin: Yes, didn't think about this. 
I will make these changes. Should I open a new PR?

smarru: yes please, can you please create a new PR and reference this PR? -- thanks.

sachinkariyattin: I have made the changes and linked the PR

PR #158, opened 2018-01-14
Fix issues with clean maven build.

syodage:
After adding this changes, both `master` and `develop` branches build success.


machristie:
Thanks @syodage ! Any idea why these changes are needed, why the build started failing now?

PR #163, opened 2018-01-22
Implemented getAllGroupsUserBelongs API method

sachinkariyattin:
The implementation _getAllMemberGroupsForUser()_ was available in the sharing registry module.


machristie:
Looks good, thanks @sachinkariyattin

machristie:
My only feedback is that I think we should be following the pattern in AiravataServerHandler where it checks a SharingRegistry service client out of a pool of clients and then checks it back in at the end of the method.  This way these connections get cleaned up.  But that is maybe worthy of being a separate issue.

sachinkariyattin:
@machristie Thanks for the review 

I had come across this while implementing the GroupManagerService. But the ThriftClientPool is currently in the airavta-api module (org.apache.airavata.api.server.util). So, instead of having the airavata-api dependency on profile-service, I decided to use a regular client. 

Suresh has already created the following Jira https://issues.apache.org/jira/browse/AIRAVATA-2607 to move the ThriftClientPool to commons module.

machristie:
Good point, thanks @sachinkariyattin . I'll link the two issues.

PR #165, opened 2018-01-29
[AIRAVATA-2601] Adding groups to Applications

tilaks26:
Filtering of Application Deployments by gateway ID and user name. We first get a list of Application Deployment IDs that are accessible by the current user and then retrieve the list of Application Deployments. (Similar to how we search for Experiments).

Files changed:

AiravataDataMigrator.java -> Included a new Entity Type names APPLICATION
AiravataServerHandler.java -> Modified getAllApplicationDeployments(), getAllAppModules() methods to filter the accessible Application IDs using the user name, modified launchExperiment() to first check if the user has access to launch an Experiment in the Application Module
registry-api.thrift -> Modified getAllApplicationDeployments(), getAllAppModules() methods' parameter list to include the accessible Application IDs
RegistryService.java -> Regenerated the Thrift stubs which changed the Registry Service interface
ApplicationDeployment.java -> Modified getAllApplicationDeployements() method to include accessible Application Deployment IDs
ApplicationInterface.java -> Modified getAllAppModules() method to include accessible Application Module IDs
RegistryServerHandler.java -> Modified getAllApplicationDeployments(), getAllAppModules() methods to reflect the changes in ApplicationDeployment.java and RegistryService.java files


tilaks26:
Made necessary changes to thrift models -
- airavata_api.thrift
- registry-api.thrift

Regenerated the services from thrift models - 
- RegistryService.java
- Airavata.java

Added new methods - 
- getAccessibleAppModules()
- getAccessibleApplicationDeployments()

Reverted the old methods - 
- getAllAppModules()
- getAllApplicationDeployments()

Made changes to the interfaces -
- ApplicationDeployment.java
- ApplicationInterface.java

modules/sharing-registry/sharing-data-migrator/src/main/java/org/apache/airavata/sharing/registry/migrator/airavata/AiravataDataMigrator.java
```diff
@@ -181,6 +189,33 @@ public static void main(String[] args) throws SQLException, ClassNotFoundExcepti
...
+        query = "SELECT * FROM APPLICATION";
+        statement = expCatConnection.createStatement();
+        rs = statement.executeQuery(query);
+        while(rs.next()){
+            try {
+                Entity entity = new Entity();
+                entity.setEntityId(rs.getString("APPLICATION_ID"));
+                entity.setDomainId(rs.getString("GATEWAY_ID"));
+                entity.setEntityTypeId(rs.getString("GATEWAY_ID") + ":APPLICATION");
+                entity.setOwnerId(rs.getString("USER_NAME") + "@" + rs.getString("GATEWAY_ID"));
```
machristie: Sneha, for the owner of an application deployment, have the migration script take an argument for a default owner that will be used to assign ownership in cases like this where there is no owner associated in the app catalog.

tilaks26: Sure, Marcus. Do you have a text in mind for default owner?

tilaks26: And what about the Entity's name?

machristie:
Sneha, this looks good. I think one additional thing we'll need to do is to share every application deployment with the "everyone" group. Can you add that to this script as well? So what we'll do is run the everyone migration that you created in the other pull request and then we can run this migration script.

tilaks26:
@machristie , do you want me to use the Sharing Client like in the everyone migrator or follow the Sharing Handler convention in the data migrator?

machristie:
@tilaks26 yeah I think it should use the sharing client too.

tilaks26:
@machristie I've made the requested changes.

PR #166, opened 2018-01-29
Fixing Airavata-2143

DImuthuUpe:


modules/gfac/gfac-core/src/main/java/org/apache/airavata/gfac/core/context/ProcessContext.java
```diff
@@ -681,7 +687,7 @@ public String getQualityOfService() {
 	public String getQueueName() {
 		if (isUseUserCRPref() &&
 				userComputeResourcePreference != null &&
-				isValid(userComputeResourcePreference.getPreferredBatchQueue())) {
+				userComputeResourcePreference.getPreferredBatchQueue() != null) {
```
machristie: Why switch from `isValid` to `!= null`? Is it because the empty string is valid here (I wouldn't think it is)?  Same question on lines 658 and 679.

modules/gfac/gfac-core/src/main/java/org/apache/airavata/gfac/core/context/ProcessContext.java
```diff
@@ -681,7 +687,7 @@ public String getQualityOfService() {
 	public String getQueueName() {
 		if (isUseUserCRPref() &&
```
machristie: Also, I don't think this logic is correct. I think processModel.getProcessResourceSchedule() should be preferred first and if it doesn't specify a queue then it should fall back to the preferences.

modules/gfac/gfac-core/src/main/java/org/apache/airavata/gfac/core/context/ProcessContext.java
DImuthuUpe: if the isUseUserCRPref() is true, we should fetch whatever we have in userComputeResourcePreference. Issue for Airavata 2143 is being some values as "" in userComputeResourcePreference and fetching them from gateway preferences. Checking for null is just a sanity check.

modules/gfac/gfac-core/src/main/java/org/apache/airavata/gfac/core/context/ProcessContext.java
DImuthuUpe: I kept it as it is as it was in the initial implementation. Will bring it into top. Thanks for pointing out.

PR #170, opened 2018-02-16
Ansible changes for standalone Airavata installation

DImuthuUpe:


dev-tools/ansible/inventories/standalone/group_vars/pga/vars.yml
```diff
@@ -0,0 +1,70 @@
...
+
+gateway_id: "default"
+# relative to document root dir
+experiment_data_dir: "{{ user_data_dir }}/exp-dir"
+gateway_data_store_resource_id: "CHANGEME"
+gateway_data_store_ssh_public_key: "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQClwPzT95sK4eA5vuOHe+9cDewjkfBZA0GLt7SfHtJ5Zp/DBWRb5ykobX7JbwJGN9eHmVnW1yL5O708/dLL3y8kT0MjEP5jxZKMSayAiP9c2yN8dved4DIomX3omqsztIcbllei+QM9YvFY7tTtjWFTtAX6WxkZEvzcv2MDTLbgqJ537B4WRRWchXFws98QfDZvlfvdql6BcoZMwDCjdNOl8ZqSTS0PC22+HFesdhMh4mjjNj3Sm2MqxjU8FmSjVlC37dubqS9lZMeKuwsgT1nMjCB8tL+FH5eneNLVKvsvJL4AgQNo84YFi/Gels40zVuCqXVRNsAw/Tn+9sH9JgVx"
+
+## Portal related variables
+super_admin_portal: "true"
+admin_emails: "['dimuthu.upeksha2@gmail.com']"
```
machristie: @DImuthuUpe  do you want your email address in here? You might end up getting a lot of spam :)

dev-tools/ansible/roles/api-orch/defaults/main.yml
```diff
@@ -28,4 +28,14 @@ iam_server_super_admin_username: "admin"
 profile_service_host: "localhost"
 profile_service_port: 8962
 
-mariadb_connector_jar_url: "https://downloads.mariadb.com/Connectors/java/connector-java-2.0.2/mariadb-java-client-2.0.2.jar"
+mariadb_connector_jar_url: "http://downloads.mariadb.com/Connectors/java/connector-java-2.0.2/mariadb-java-client-2.0.2.jar"
```
machristie: Why was this changed to "http"?

dev-tools/ansible/roles/env_setup/tasks/main.yml
```diff
@@ -53,4 +53,8 @@
   service: name=firewalld state=started
   become: yes
 
+- name: open firewall port 22 for SSH connections
```
machristie: Is this needed? If port 22 isn't open how will Ansible be able to connect to the host?

dev-tools/ansible/roles/gfac/defaults/main.yml
```diff
@@ -19,4 +19,7 @@
 #
 
 gfac_log_dir: "{{ gfac_dir }}/{{ airavata_dist }}/logs"
-mariadb_connector_jar_url: "https://downloads.mariadb.com/Connectors/java/connector-java-2.0.2/mariadb-java-client-2.0.2.jar"
+mariadb_connector_jar_url: "http://downloads.mariadb.com/Connectors/java/connector-java-2.0.2/mariadb-java-client-2.0.2.jar"
```
machristie: Same question as above about "http"

modules/credential-store/credential-store-service/src/main/java/org/apache/airavata/credential/store/store/impl/util/CredentialStoreInitUtil.java
```diff
@@ -86,6 +98,35 @@ public static void initializeDB() {
                 logger.error("Error while closing database connection...", e.getMessage(), e);
             }
         }
+
+        try {
+
+            GatewayProfileResource gatewayProfileResource = new GatewayProfileResource();
```
machristie: This should use the Registry API rather than the Resource classes directly.

But I don't think this is the best way to do it. The credential store just stores credentials, it doesn't know anything about the default gateway or the app catalog.  I think the following two places are better:
* https://github.com/apache/airavata/blob/master/modules/registry/registry-server/registry-api-service/src/main/java/org/apache/airavata/registry/api/service/util/AppCatalogInitUtil.java#L101
  * however, this might not work since the credential store is started after the registry, see ServerMain.java
* when the API server starts, [similar to how the sharing registry is initialized](https://github.com/apache/airavata/blob/master/airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java#L122)

dev-tools/ansible/roles/rabbitmq/tasks/main.yml
```diff
@@ -32,7 +32,8 @@
   # add hostname to /etc/hosts file
 - name: get ip4 address
   # command: dig +short myip.opendns.com @resolver1.opendns.com
-  command: hostname -i
+  # select the first ip address when there are more than one
+  shell: hostname -i | tr -s ' ' | cut -d ' ' -f 1
```
machristie: Can we use the ansible variable `ansible_default_ipv4.address` instead?

dev-tools/ansible/roles/pga/templates/pga-ssl-vhost.conf.j2
```diff
@@ -1,4 +1,4 @@
-<VirtualHost *:80>
+<VirtualHost *:8008>
```
machristie: Shouldn't these files use the `pga_default_http_port` variables? See also line 13 and pga-vhost.conf.j2.

dev-tools/ansible/inventories/standalone/group_vars/pga/vars.yml
DImuthuUpe: My bad. I will remove the email from here

dev-tools/ansible/roles/api-orch/defaults/main.yml
DImuthuUpe: MariaDB ssl certificate was expired in the day I was testing this. Now it seems like fixed. Will revert back to https now

dev-tools/ansible/roles/env_setup/tasks/main.yml
DImuthuUpe: Assume the machine do not have firewalld installed at the beginning and then we install firewalld on this. By default firewalld blocks all the ports. So it blocks next ssh connections if we do not whitelist port 22 in the same connection that installed firewalld.

dev-tools/ansible/roles/gfac/defaults/main.yml
DImuthuUpe: Same reply as above

modules/credential-store/credential-store-service/src/main/java/org/apache/airavata/credential/store/store/impl/util/CredentialStoreInitUtil.java
DImuthuUpe: I also initially thought of of putting it to AppCatalogInitUtil but due to the initialization order, it didn't work. I'll move the code to API server which makes more sense.

dev-tools/ansible/roles/api-orch/defaults/main.yml
DImuthuUpe: I again reverted back to http urls as ansible doesn't trust the ss certificates provided by mariadb

fatal: [192.168.99.102]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to validate the SSL certificate for downloads.mariadb.com:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine or you can install the `urllib3`, `pyopenssl`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible"}

machristie: That's odd, I wonder why this hasn't been a problem before.  Maybe the version of Python on the VM you're testing with?  In any case, I'm fine with this change.

dev-tools/ansible/roles/env_setup/tasks/main.yml
machristie: Yes, this makes sense. I was thinking of the OpenStack firewall. However, why have we never run into this issue previously?

machristie:
Looks good @DImuthuUpe . I'm going to merge this into the develop branch and run a deploy just to test it out first before merging to the master branch.

dev-tools/ansible/roles/env_setup/tasks/main.yml
DImuthuUpe: I tested in Virtualbox and I had to manually whitelist port 22 if it is not being done at this step.

PR #172, opened 2018-02-21
[AIRAVATA-2679]  [AIRAVATA-2680] GroupManagerService: Auth checks and add/remove users from group

sachinkariyattin:
**Changes**:

- Added two API methods to the GroupMangerService `addUsersToGroup()` and `removeUsersFromGroup`
- Added necessary auth checks for owner and admin


machristie:
@sachinkariyattin This looks great! Thanks for quickly putting this together. 

It does look like a couple of sharing tests failed:
* org.apache.airavata.sharing.registry.SharingRegistryServiceTest
* org.apache.airavata.sharing.registry.SharingRegistryServerHandlerTest

They both fail with this error
```
org.apache.airavata.sharing.registry.models.SharingRegistryException: New group owner is not part of the group Stack trace:SharingRegistryException(message:New group owner is not part of the group)
```

I wonder if maybe transferGroupOwnership should just add the new owner to the group if the new owner isn't already a member of the group?

sachinkariyattin:
@machristie Thanks for the review. I forgot to update the tests. Fixed it now.

PR #174, opened 2018-02-21
Registry Refactoring : ComputeResourceEntity fix

sachinkariyattin:
Changes:

- `ComputeResourceEntity.gatewayUsageReporting` to boolean instead of boolean


machristie:
Thanks @sachinkariyattin . I think we'll also need this for all of the boolean fields in the entities.

sachinkariyattin:
@machristie I will consider this and make changes to all the entities in my next PR for registry refactoring

PR #178, opened 2018-02-27
[AIRAVATA-2699] Implement Group resource profile validation when launching an experiment

sachinkariyattin:
Implemented GroupResourceProfile validator to validate maxNodes, maxCores and maxWallTime. 

**Changes**:

- Added _groupResourceId_ field in the UserConfigurationDataModel.
- Implemented _GroupResourceProfileValidator_ class (_orchestrator->validator->_GroupResourceProfileValidator__)
- Added a new column _groupResourceId_ to the UserConfiguration table as well(sql changes)


modules/orchestrator/orchestrator-core/src/main/java/org/apache/airavata/orchestrator/core/validator/impl/GroupResourceProfileValidator.java
```diff
@@ -0,0 +1,155 @@
...
+
+        String groupResourceProfileId = userConfigurationData.getGroupResourceProfileId();
+        String computeResourceId;
+        if (processModel == null) {
+            computeResourceId = computationalResourceScheduling.getResourceHostId();
+        } else {
+            computeResourceId = processModel.getProcessResourceSchedule().getResourceHostId();
+        }
+
+        List<BatchQueueResourcePolicy> batchQueueResourcePolicies = registryClient.getGroupBatchQueueResourcePolicyList(groupResourceProfileId);
```
machristie: It's possible for a GroupResourceProfile to have no ComputeResourcePolicy or BatchQueueResourcePolicies. So this logic needs to change slightly so that it only enforces those policies if they exist.

modules/orchestrator/orchestrator-core/src/main/java/org/apache/airavata/orchestrator/core/validator/impl/GroupResourceProfileValidator.java
```diff
@@ -0,0 +1,155 @@
...
+
+                    ValidatorResult cpuCountResult = new ValidatorResult();
+
+                    if (experimentCPUCount > batchQueueResourcePolicy.getMaxAllowedCores()) {
+                        cpuCountResult.setResult(false);
+                        cpuCountResult.setErrorDetails("Job Execution cpu count " + experimentCPUCount +
+                                "exceeds the allowable cpu count for the group resource profile" + batchQueueResourcePolicy.getMaxAllowedCores() +
+                                "for queue " + batchQueueResourcePolicy.getQueuename());
+                    } else {
+                        cpuCountResult.setResult(false);
```
machristie: result should be true

modules/orchestrator/orchestrator-core/src/main/java/org/apache/airavata/orchestrator/core/validator/impl/GroupResourceProfileValidator.java
```diff
@@ -0,0 +1,155 @@
...
+            } else {
+                queueNameResult.setResult(false);
+                queueNameResult.setErrorDetails("The specified queue " + queueName +
+                        " does not exist in the list of allowed queues for the group resource profile.");
+                validatorResultList.add(queueNameResult);
+            }
+        } else {
+            ValidatorResult result = new ValidatorResult();
+            logger.info("There is no compute resource policy specified for the group resource profile");
+            result.setResult(true);
```
machristie: result should be false, but also, as mentioned above, we don't want to invalidate if there is no compute resource policy.

modules/orchestrator/orchestrator-service/src/main/java/org/apache/airavata/orchestrator/server/OrchestratorServerHandler.java
```diff
@@ -146,10 +148,20 @@ public boolean launchExperiment(String experimentId, String gatewayId) throws TE
...
+							userConfigurationData.getComputationalResourceScheduling().getResourceHostId());
             String token = computeResourcePreference.getResourceSpecificCredentialStoreToken();
+
+			if (userConfigurationData.getGroupResourceProfileId() != null) {
+				GroupComputeResourcePreference groupComputeResourcePreference = registryClient.getGroupComputeResourcePreference(
+						userConfigurationData.getComputationalResourceScheduling().getResourceHostId(),
+						userConfigurationData.getGroupResourceProfileId());
+				if (groupComputeResourcePreference.getResourceSpecificCredentialStoreToken() != null) {
+					token = groupComputeResourcePreference.getResourceSpecificCredentialStoreToken();
+				}
```
machristie: If there is no resource specific credential store token on the groupComputeResourcePreference, it should then take the resource specific credential store token from the gateway computeResourcePreference.

modules/orchestrator/orchestrator-core/src/main/java/org/apache/airavata/orchestrator/core/validator/impl/GroupResourceProfileValidator.java
machristie: As we discussed, I misread this, you're right the result should be true.

modules/orchestrator/orchestrator-service/src/main/java/org/apache/airavata/orchestrator/server/OrchestratorServerHandler.java
machristie: Also as we discussed, you're right about this one too. The code is correctly defaulting to the compute pref resource specific credential store token and overriding that if there is one defined for the GroupResourceProfile's compute pref.

sachinkariyattin:
@machristie Thanks for the review. I have implemented the changes. Please have a look.

modules/orchestrator/orchestrator-core/src/main/java/org/apache/airavata/orchestrator/core/validator/impl/GroupResourceProfileValidator.java
```diff
@@ -0,0 +1,182 @@
...
+        int experimentWallTimeLimit = computationalResourceScheduling.getWallTimeLimit();
+        int experimentNodeCount = computationalResourceScheduling.getNodeCount();
+        int experimentCPUCount = computationalResourceScheduling.getTotalCPUCount();
+
+        ValidatorResult wallTimeResult = new ValidatorResult();
+
+        if (experimentWallTimeLimit > batchQueueResourcePolicy.getMaxAllowedWalltime()) {
+            wallTimeResult.setResult(false);
+            wallTimeResult.setErrorDetails("Job Execution walltime " + experimentWallTimeLimit +
+                    "exceeds the allowable walltime for the group resource profile" + batchQueueResourcePolicy.getMaxAllowedWalltime() +
```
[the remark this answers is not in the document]
sachinkariyattin: Done!

PR #179, opened 2018-03-01
[AIRAVATA-2696] Add GroupResourceProfile to ProcessModel

sachinkariyattin:
Added GroupResourceProfile related changes to gfac module

**Changes:**

- Added groupResourceProfileId to _ProcessModel_.
- Added _GroupResourceProfile_ and _GroupResourcePreference_ objects to _ProcessContext_. ComputeResource credential token will be loaded from the group compute preference if it exists otherwise the credential token is loaded from GatewayComputeResourcePreference.
- Added new column(GROUP_RESOURCE_PROFILE_ID) to the PROCESS table (sql script changes).


machristie:
Looks really good Sachin! Thanks!

PR #180, opened 2018-03-08
Fixing the document root of PGA installation Ansible script

Lahiru-J:
Existing DocumentRoot `/www/default` will lead to Apache Server Testing page when entering 
`http://<pga-host>:8008/` to configure PGA. Correct DocumentRoot should be `/var/www/portals/default/public`

### DIFF
--- a/dev-tools/ansible/roles/pga/templates/default.conf.j2
+++ b/dev-tools/ansible/roles/pga/templates/default.conf.j2
@@ -2,5 +2,5 @@
 # This will show the welcome page when requesting on the ip address or server default hostname
 Listen {{ pga_default_http_port }}
 <VirtualHost _default_:{{ pga_default_http_port }}>
-    DocumentRoot "/www/default"
+    DocumentRoot {{ doc_root_dir }}/public
 </VirtualHost>


machristie:
This won't work when deploying multiple PGAs in the same Apache server, they will each overwrite the other.  This default configuration was added to prevent Apache from displaying some random virtual host when the IP address or some non-virtual host domain name is visited.

machristie:
Why didn't the virtual host configuration work in pga-default.conf?

Lahiru-J:
The format of `<VirtualHost *:8008>` in `pga-default.conf` didn't work for me. I had to manually change it to `<VirtualHost 192.168.1.10:8008>` (192.168.1.10 - my IP address)

machristie:
So going to http://192.168.1.10:8008 didn't load the PGA?  You got the default Apache page?  Can you dig into this a little more and find out why going to http://192.168.1.10:8008 didn't work?

Lahiru-J:
yes @machristie it didn't work, I always got the Apache Testing page. Sure I will try to find it.

Lahiru-J:
Hi @machristie, 
The problem was server doesn't pick the VirtualHost as the best match for the particular IP address and port. This [answer](The problem was server doesn't pick the VirtualHost as the best match for the particular IP address and port. ) confirms it. 
There were some solutions in order to get the work done

1. Change `<VirtualHost *:8008>` into `<VirtualHost 192.168.1.10:8008>` in **pga-default.conf**
2. Add a ServerName to default VirtualHost in **default.conf**

But the problem was, with the configuration `<VirtualHost *:8008>` some of my colleagues have received successful results. This problem occurred only for me. This [link](https://serverfault.com/questions/901342/virtualhost-is-not-taken-as-the-best-match?noredirect=1#comment1165724_901342) is the answer for that (I question that I asked in **serverfault**

PR #184, opened 2018-03-16
Fixing Airavata 2625 : Derive and present Text outputs in Experiment summary

DImuthuUpe:
For more details, refer to https://issues.apache.org/jira/browse/AIRAVATA-2625


modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
```diff
@@ -111,24 +136,162 @@ public TaskStatus execute(TaskContext taskContext) {
...
+
+                log.info("Downloading file " + sourceURI.getPath() + " to temporary file " + tempFile);
+
+                taskContext.getParentProcessContext().getJobSubmissionRemoteCluster().copyFrom(sourceURI.getPath(), tempFile.getPath());
+
+                StringBuilder result = new StringBuilder();
+                // this is to identify that output is parsed
+                result.append("parsed-out: ");
+
+                // if the file size is grater than 2 MB, skip parsing to avoid possible OOM issues
```
machristie: I don't think the OOM issue is a concern. The code isn't reading the entire file into memory. The BufferedReader only uses a relatively small buffer for reading the file.

One file size concern though is that if it is very large it may fill up the disk. If I'm understanding the code correctly this transfers the file from the compute resource to a temp file (in /tmp?) on the gfac server.  We generally don't provision a lot of storage space on gfac so /tmp may only have a few GBs, but some output files can be very large.  That would be a misconfiguration, but it would also disrupt service.

However, maybe not worth trying to solve at this time. I guess you could run something over SSH to check the size of the file before transferring it.

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
```diff
@@ -111,24 +136,162 @@ public TaskStatus execute(TaskContext taskContext) {
...
+                        while ((line = br.readLine()) != null) {
+
+                            if (log.isTraceEnabled()) {
+                                log.trace("Input file line : {}", line);
+                            }
+
+                            for (String arg : allArgs) {
+                                if (line.startsWith(arg)) {
+                                    log.debug("Found a line with argument {} : {}", arg, line);
+                                    result.append(line).append("\n");
```
machristie: Are multiple values supported? Or should this break after finding the first one?

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
```diff
@@ -97,7 +100,29 @@ public TaskStatus execute(TaskContext taskContext) {
...
+
+                }
+
+                if (processOutput.getType() == DataType.URI || processOutput.getType() == DataType.URI_COLLECTION ||
+                        processOutput.getType() == DataType.STDOUT || processOutput.getType() == DataType.STDERR) {
+                    return transferFiles(processContext, subTaskModel, taskContext, processState);
+
+                } else if (processOutput.getType() == DataType.FLOAT || processOutput.getType() == DataType.STRING ||
+                        processOutput.getType() == DataType.INTEGER ) {
+                    return extractStringFromFile(subTaskModel, taskContext, processOutput.getApplicationArgument());
```
machristie: This seems like a bit of a hack to use applicationArgument this way. Would it not be better to add another property to the OutputDataObjectType model?

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
DImuthuUpe: Agreed. I misjudged OOM issue here. Thanks for pointing that out. Will remove this check and do a size check before transferring the file.

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
DImuthuUpe: Do you suggest to add something like readable/ transferrable?

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
```diff
@@ -51,42 +47,49 @@
...
-    private static final int DEFAULT_SSH_PORT = 22;
+
+    /**
+     * Maximum size of the file which is compatible for STRING output parsing
+     */
+    private static final long MAX_FILE_SIZE_TO_READ = 2 * 1024 * 1024;
 
     @Override
-    public void init(Map<String, String> propertyMap) throws TaskException {
+    public void init(Map<String, String> propertyMap) {
```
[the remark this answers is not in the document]
DImuthuUpe: I removed this as this overridden method does not actually return a TaskException. However now I think better to bring it back to have the consistency in the code. Thanks for pointing out

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
machristie: I was thinking of something like "outputValuePrefix" to define the string to look for in the file.

machristie:
@DImuthuUpe Just a general comment: I'm not sure if SCPDataStageTask is the best place for this. Probably it is the best place on master, but with the Helix task execution implementation I think this logic would be better in a separate task. What do you think?

DImuthuUpe:
@machristie In helix this is a separate task.

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
DImuthuUpe: What could be possible values for outputValuePrefix? And on what condition do you suggest to trigger data extraction?

DImuthuUpe:
@machristie I updated the PR to read file size remotely and then download the file. However I did not try to address that output type filtering improvement as we ultimately discard this code and move into Helix based one and we can implement this in a much cleaner manner.

modules/gfac/gfac-impl/src/main/java/org/apache/airavata/gfac/impl/task/SCPDataStageTask.java
machristie: You would keep triggering data extraction based on the output type (FLOAT, STRING, INTEGER). But the string you look for would be specified by outputValuePrefix. Possible values would be null or some string, and it should not be null if the output type is FLOAT, STRING or INTEGER.

The reason I don't think we should use applicationArgument is because it has a different use, it is used to determine the command line, see https://github.com/apache/airavata/blob/433800846fcaa9d254fe598af0958fec9431cc75/modules/gfac/gfac-core/src/main/java/org/apache/airavata/gfac/core/GFacUtils.java#L648-L648.

machristie:
@DImuthuUpe the changes look good.

Overall, the code looks good and I can see why this feature is useful, but I have some reservations about this approach and what it means for the data model.  I don't think we should have OutputDataObjectType have DataTypes of STRING, INTEGER or FLOAT. Those DataType only make sense for application inputs, they don't make sense for application outputs. For Airavata, application outputs are only files. Extracting a scalar value from an application output file should be done by a downstream consumer in a workflow instead of directly in the application execution.

However, I'm fine with moving forward with this as a short term solution because, as I said, I can see the value of it.

PR #189, opened 2018-04-04
Using one RabbitMQ Channel for one thread

DImuthuUpe:
Currently, using a RabbitMQ publisher multiple threads can publish to the broker through a single channel. However RabbitMQ mentions not to reuse same channel among multiple threads as it is not thread safe. So I used a thread local to store the channel object and each thread will use a separate channel object.


DImuthuUpe:
@machristie Good question. We might have to add a close method to Publisher Interface and register it in a shutdown hook if we are not going to manually close it

PR #190, opened 2018-04-07
Fixing output files not staging issues when wildcards are specified

DImuthuUpe:


DImuthuUpe:
@machristie Thanks for the review. This works for multiple files as well. Have a look at line 197
processOutput.setName(fileName);
Output name is changed for each fetched file name.

DImuthuUpe:
Merging this and changing the same in master branch

machristie:
@DImuthuUpe Thanks, yes I see, so a new output is registered for each file name. I still think we need to think about this some more. It makes sense to me that we would register multiple files matching a wildcard as a COLLECTION of files.

PR #199, opened 2018-05-21
Fixed the table column constraint

rkunalan:
In table "USER_COMPUTE_RESOURCE_PREFERENCE", "SCRATCH_LOCATION" column contained the conflicting constraint. Fixed that column constraint.


DImuthuUpe:
@rkunalan Thanks for the PR. 
@machristie Do are you ok with this change? Are we expecting SCRATCH_LOCATION to be a null or not null field? I think this issue exists in other branches as well (develop, staging)

machristie:
@DImuthuUpe I think this is ok and it is consistent with the other fields on the table and the JPA entity. Truthfully it should be NOT NULL, but the user interface is enforcing that.

PR #203, opened 2018-07-24
Adding Workflow Related Thrift Models and Database Models

yasgun:
- Modifying API to accept workflow submissions
- Adding workflow Thrift models
- Adding JPA entity classes
- Adding workflow repository
- Adding test cases


modules/registry-refactoring/src/main/java/org/apache/airavata/registry/core/repositories/workflowcatalog/WorkflowRepository.java
```diff
@@ -17,16 +17,13 @@
...
 package org.apache.airavata.registry.core.repositories.workflowcatalog;
 
-import org.apache.airavata.model.WorkflowModel;
-import org.apache.airavata.model.application.io.OutputDataObjectType;
-import org.apache.airavata.registry.core.entities.workflowcatalog.WorkflowEntity;
-import org.apache.airavata.registry.core.utils.DBConstants;
-import org.apache.airavata.registry.core.utils.ObjectMapperSingleton;
-import org.apache.airavata.registry.core.utils.QueryConstants;
-import org.apache.airavata.registry.core.utils.WorkflowCatalogUtils;
+import org.apache.airavata.model.commons.airavata_commonsConstants;
```
DImuthuUpe: Why not AiravataCommonsConstants ?

modules/registry-refactoring/src/main/java/org/apache/airavata/registry/core/repositories/workflowcatalog/WorkflowRepository.java
```diff
@@ -39,117 +36,159 @@
...
+    protected AiravataWorkflowEntity saveWorkflow(AiravataWorkflow workflowModel, String gatewayId) throws WorkflowCatalogException {
+
+        if (workflowModel.getId() == null || workflowModel.getId().equals(airavata_commonsConstants.DEFAULT_ID)) {
+            logger.debug("Setting the ID for the new Workflow");
+            workflowModel.setId(WorkflowCatalogUtils.getID(workflowModel.getName()));
         }
 
-        String templateId = workflowModel.getTemplateId();
+        if (workflowModel.getStatuses() != null) {
+            logger.debug("Populating the status ID of WorkflowStatus objects for the Workflow");
```
DImuthuUpe: Print the workflow id as well. Otherwise there is no use of this log line

modules/registry-refactoring/src/main/java/org/apache/airavata/registry/core/repositories/workflowcatalog/WorkflowRepository.java
```diff
@@ -39,117 +36,159 @@
...
     }
 
-    protected WorkflowEntity saveWorkflow(WorkflowModel workflowModel, String gatewayId) throws WorkflowCatalogException {
-        if (workflowModel.getTemplateId() == null) {
-            logger.debug("Setting the Template ID for the new Workflow");
-            workflowModel.setTemplateId(WorkflowCatalogUtils.getID(workflowModel.getName()));
+    protected AiravataWorkflowEntity saveWorkflow(AiravataWorkflow workflowModel, String gatewayId) throws WorkflowCatalogException {
+
+        if (workflowModel.getId() == null || workflowModel.getId().equals(airavata_commonsConstants.DEFAULT_ID)) {
+            logger.debug("Setting the ID for the new Workflow");
```
DImuthuUpe: Print applied workflow id

modules/registry-refactoring/src/main/java/org/apache/airavata/registry/core/repositories/workflowcatalog/WorkflowRepository.java
```diff
@@ -39,117 +36,159 @@
...
+            logger.debug("Populating the status ID of WorkflowStatus objects for the Workflow");
+            workflowModel.getStatuses().forEach(workflowStatus -> {
+                if (workflowStatus.getId() == null) {
+                    workflowStatus.setId(WorkflowCatalogUtils.getID("WORKFLOW_STATUS"));
+                }
+            });
+        }
+
+        if (workflowModel.getGatewayId() == null) {
+            logger.debug("Setting the GatewayID for the new Workflow");
```
DImuthuUpe: Refer to previous comments

modules/registry-refactoring/src/main/java/org/apache/airavata/registry/core/repositories/workflowcatalog/WorkflowRepository.java
```diff
@@ -39,117 +36,159 @@
...
+        if (workflowEntity.getNotificationEmails() != null) {
+            logger.debug("Populating the Workflow ID of NotificationEmail objects for the Workflow");
+            workflowEntity.getNotificationEmails().forEach(notificationEmailEntity -> notificationEmailEntity.setWorkflowId(workflowId));
+        }
 
-        if (gatewayId != null) {
-            logger.debug("Setting the gateway ID of the Workflow");
-            workflowEntity.setGatewayId(gatewayId);
+        if (workflowEntity.getStatuses() != null) {
+            logger.debug("Populating the Workflow ID of WorkflowStatus objects for the Workflow");
```
DImuthuUpe: Refer to previous comments

DImuthuUpe:
@yasgun looks good to me. Please have a look at the comments. They are very minor changes though

modules/registry-refactoring/src/main/java/org/apache/airavata/registry/core/repositories/workflowcatalog/WorkflowRepository.java
yasgun: @DImuthuUpe It's an old thrift auto generated file. Need to change the .thrift filename along with few modifications to the related usages, in order to fix this. Is it OK to do it in another PR?

DImuthuUpe: @yasgun It's better to send it as a separate PR however let's keep it as it is for the instance.

DImuthuUpe:
@yasgun after merging your thrift interface modifications, thrift python client stopped working. Here is a sample python code that can reproduce the issue.

import sys
sys.path.append('/Users/dimuthu/code/reference/airavata/thrift-interface-descriptions/target/gen-py')
from airavata.api import Airavata
from thrift import Thrift
from thrift.transport import TSocket, TSSLSocket
from thrift.transport import TTransport
from thrift.protocol import TBinaryProtocol
from airavata.model.security.ttypes import AuthzToken

Here is the error that I'm seeing https://gist.github.com/DImuthuUpe/fb7e9654c3c32edf2ccdfd36943e7156. 

For the instance, I'm accumulating all the workflow related thrift files into a single file to make it work but you might need to refactor them in order to preserve modularity. And I'm removing handlers field from WorkflowHandler struct as in python you can't refer same class inside the same class. You might have to figure out a better approach to fix that.

DImuthuUpe:
Temporary fixes can be found from this change log https://github.com/apache/airavata/commit/b377c7dff620d1286f976ccf3f15ea203dd07702

machristie:
@DImuthuUpe ,

I'm not sure what is causing this error but my guess is it is caused by the [other workflow thrift file which generates python code in the same namespace](https://github.com/apache/airavata/blob/3451a63dd21c96978226bc30754e94196a4519c3/thrift-interface-descriptions/data-models/workflow-models/workflow_data_model.thrift#L24-L24), which will surely cause conflicts.  Probably it would be best to just remove the old workflow thrift files since we aren't using them.

DImuthuUpe:
@machristie That could be one reason and he had multiple thrift files pointing to same namespace and for some reason only one file is picked by thrift compiler. I merged all the files into a single file and it worked.

yasgun:
If we cannot have multiple Thrift files pointing to the same namespace for python compilation, I think we have to forget about preserving modularity for the moment and keep it in a single file. (Which looks OK to me :+1:  ) But for WorkflowHandler I think there is no option than referring to the same class within it. @DImuthuUpe Can we use metaclasses (or other workaround) and implement that manually for Python client if Thrift doesn't support this? Or any other suggestion to overcome this?

PR #216, opened 2019-03-15
Fixing case sensitive issues in database initialization script.

sairohithA007:


DImuthuUpe:
Thanks @sairohithA007

PR #217, opened 2019-03-15
updated README.md

Keerthi4308:
I added direct link to npm(https://www.npmjs.com/get-npm) so that if folks need to update there is a quick command to run along with download option.


machristie:
Looks good to me. Thanks Keerthi!

PR #218, opened 2019-04-12
Change Date.getTime() to System.currentTimeMillis()

bd2019us:
Hello,

new Date() is a wrapper of method System.currentTimeMillis(). If it is intensively invoked in the program, the performance will be greatly damaged.  
According to my local testing, when these two methods are invoked 5,000,000 times at the same environment, System.currentTimeMillis() can achieve a speedup to 5 times (435 ms vs 2073 ms). 
Therefore, if only getTime() is used for Date object, the light method System.currentTimeMillis() is highly recommended, which can also avoid creating the temporary Date object.

PR #220, opened 2019-04-20
use System.currentTimeMillis()

bd2019us:
same pr as #218 to use System.currentTimeMillis()


DImuthuUpe:
@bd2019us Thanks for pointing this out. Can you check whether this is the only place where you can see new Date()).getTime() in the whole code base? If not please update those areas as well to preserve consistency.

bd2019us:
No problem, I will check and update all similar locations.

DImuthuUpe:
Looks good to me

PR #221, opened 2019-05-21
Fixing two issues encountered during the setup of development environment in Ubuntu OS

shivamriky:
_modified:   modules/ide-integration/src/main/resources/airavata-server.properties_
Updated changed the server.host from localhost to 0.0.0.0 as the localhost as localhost was creating problems in Linux where
the docker container was not able to talk to the API services due to a different host address.

_modified:   pom.xml_
1. Added maven repository as maven was not able to download some libraries from already available repos.
2. Added a plugin for openJPA to avoid the following error in Intellij:
"<openjpa-2.4.3 nonfatal user error> org.apache.openjpa.persistence.ArgumentException: This configuration disallows runtime optimization, but the following listed types were not enhanced at build time or at class load time with a javaagent: "

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -553,6 +553,11 @@
             </snapshots>
         </repository>
         <repository>
+            <id>wso2</id>
+            <name>Repository</name>
+            <url>http://dist.wso2.org/maven2</url>
+        </repository>
+        <repository>
             <id>oauth</id>
             <name>OAuth Repository</name>
             <url>https://oauth.googlecode.com/svn/code/maven</url>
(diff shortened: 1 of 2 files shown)


shivamriky:
Just for reference, the following file contains the error I was getting before adding the new dependency. It also includes the successful build message after adding the new dependency. 
[maven_build_error.txt](https://github.com/apache/airavata/files/3204071/maven_build_error.txt)

DImuthuUpe:
Thanks @shivamriky. I will merge this

pom.xml
```diff
@@ -552,6 +552,11 @@
                 <enabled>false</enabled>
             </snapshots>
         </repository>
+        <repository>
+            <id>wso2</id>
+            <name>Repository</name>
+            <url>http://dist.wso2.org/maven2</url>
```
machristie: Hi @shivamriky,

Can you change this url to use `https` instead?

machristie: I went ahead and made this change.

shivamriky: Thank you. I didn't see the message earlier.

PR #225, opened 2019-05-28
Improvement for Jira 1870 - provide ability to search experiments by jobId in the experiment statistics tab under admin dashboard

shivamriky:
I have added functionality in the PGA to search experiments using the jobId. It uses an existing API function searchExperiment() to first search the experiments using a new field 'JobId'. This new field is added to the experiment_model.thrift file.
After searching the corresponding experimentid using jobid, the normal flow of information follows i.e, the experiment Id is used to search a detailed view of the experiment and then presented to the user. 

The below screenshot displays the functionality. 

image

### DIFF
--- a/modules/ide-integration/src/main/resources/pga/airavata-php-gateway/app/controllers/ExperimentController.php
+++ b/modules/ide-integration/src/main/resources/pga/airavata-php-gateway/app/controllers/ExperimentController.php
@@ -128,9 +128,32 @@ class ExperimentController extends BaseController
 
     public function summary()
     {
+        Log::info("Inside the summary function...");
+
+        Log::info(print_r($_GET, TRUE));
+
         $experiment = null;
+        $experiment_id=null;
+        if(isset($_GET['jobId'])){
+            Log::info("Summary is called");
+            $limit = 1;
+            $pageNo = 0;
+            $inputs["search-key"] = "jobId";
+            $inputs["search-value"] = $_GET['jobId'];
+            $expContainer = ExperimentUtilities::get_expsearch_results_with_pagination($inputs, $limit,
+                $pageNo);
+            Log::info("Experiment ID for the jobId is: ");
+            Log::info($expContainer[0]["experiment"]->experimentId);
+
+            $experiment_id=$expContainer[0]['experiment']->experimentId;
+        }
+        else {
+            $experiment_id = $_GET['expId'];
+        }
+
         try {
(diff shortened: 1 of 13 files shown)


modules/ide-integration/src/main/resources/pga/airavata-php-gateway/app/controllers/ExperimentController.php
```diff
@@ -129,8 +129,42 @@ public function createSubmit()
...
+            $limit = 1;
+            $pageNo = 0;
+            $inputs["search-key"] = "jobId";
+            $inputs["search-value"] = $_GET['jobId'];
+            try {
+                $expContainer = ExperimentUtilities::get_expsearch_results_with_pagination($inputs, $limit,
+                    $pageNo);
+            } catch (AuthorizationException $ae) {
+
+                Log::error("Experiment wasn't found", array("message" => $enf->getMessage(), "username" => Session::get("username"), "gateway_id" => Session::get("gateway_id")));
```
DImuthuUpe: Change the error message. This is an authorization error

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/repositories/expcatalog/ExperimentSummaryRepository.java
```diff
@@ -51,6 +56,17 @@
             throw new RegistryException("GatewayId is required");
         }
 
+        if (filters.get(DBConstants.Job.JOB_ID) != null) {
+            logger.debug("Filter Experiments by JobId");
+            queryParameters.put(DBConstants.Job.JOB_ID, filters.get(DBConstants.Job.JOB_ID));
+            String query_jobId = "SELECT P.experimentId FROM "
```
DImuthuUpe: Don't you need to filter by the gateway id as well?

shivamriky: As discussed, the Airavata API is filtering the experiments that are accessible by the user using the sharing service.

modules/ide-integration/src/main/resources/pga/airavata-php-gateway/app/controllers/ExperimentController.php
shivamriky: I have fixed it.

PR #226, opened 2019-05-31
fixed failing test cases

aarushiibisht:
UserResourceProfileRepositoryTest
The test case was failing due to failing assert statement. userResourceProfileRepository has 2 user resource profiles. A default user resource profile is added during appCatalog DB Initialization, 2nd user profile is added during testing.

SharingRegistryServerHandlerTest
Test case was failing due to failure to initialize sharing registry db. The sharingRegistryDBInitConfig being created was not being used during the creation of SharingRegistryServerHandler.

### DIFF
--- a/modules/registry/registry-core/src/test/java/org/apache/airavata/registry/core/repositories/appcatalog/UserResourceProfileRepositoryTest.java
+++ b/modules/registry/registry-core/src/test/java/org/apache/airavata/registry/core/repositories/appcatalog/UserResourceProfileRepositoryTest.java
@@ -68,7 +68,7 @@ public class UserResourceProfileRepositoryTest extends TestBase {
                 userId, gatewayId, userStoragePreference.getStorageResourceId());
         assertEquals(userStoragePreference.getFileSystemRootLocation(), retrievedUserStoragePreference.getFileSystemRootLocation());
 
-        assertTrue(userResourceProfileRepository.getAllUserResourceProfiles().size() == 1);
+        assertTrue(userResourceProfileRepository.getAllUserResourceProfiles().size() == 2);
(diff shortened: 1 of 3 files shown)


modules/registry/registry-core/src/test/java/org/apache/airavata/registry/core/repositories/appcatalog/UserResourceProfileRepositoryTest.java
```diff
@@ -68,7 +68,7 @@ public void UserResourceProfileRepositoryTest() throws AppCatalogException {
                 userId, gatewayId, userStoragePreference.getStorageResourceId());
         assertEquals(userStoragePreference.getFileSystemRootLocation(), retrievedUserStoragePreference.getFileSystemRootLocation());
 
-        assertTrue(userResourceProfileRepository.getAllUserResourceProfiles().size() == 1);
+        assertTrue(userResourceProfileRepository.getAllUserResourceProfiles().size() == 2);
```
machristie: Before we make this change, can you check with @DImuthuUpe whether we still need to create the UserResourceProfile in the AppCatalogDBInitConfig? https://github.com/apache/airavata/blob/develop/modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/utils/AppCatalogDBInitConfig.java#L70-L77

If we don't need the UserResourceProfile I'd prefer that we remove it from AppCatalogDBInitConfig instead of changing this test.

aarushiibisht: removed the additional user being created during db initialization and reverted the test case after confirming with @DImuthuUpe. All test cases are passing now. Tested the login functionality in Django portal using user default-admin.

machristie:
Thanks @aarushiibisht !

PR #229, opened 2019-07-12
refactored the create group method to add admins in the group

aarushiibisht:
Fixed bug https://issues.apache.org/jira/browse/AIRAVATA-3111


machristie:
Thanks @aarushiibisht

PR #235, opened 2019-10-23
Refactored airavatadatamigrator to add missing active keycloak users to sharing registry and groups

aarushiibisht:


modules/sharing-registry/sharing-data-migrator/src/main/java/org/apache/airavata/sharing/registry/migrator/airavata/AiravataDataMigrator.java
```diff
@@ -432,6 +444,68 @@ private static void shareEntityWithAdminGatewayGroups(SharingRegistryServerHandl
...
+
+    private static boolean migrateKeycloakUsersToGateway(IamAdminServices.Client adminServiceClient,AuthzToken authzToken, List<UserProfile> missingUsers) throws TException{
+
+        boolean allUsersUpdated = true;
+        for(UserProfile profile: missingUsers) {
+                allUsersUpdated &= adminServiceClient.enableUser(authzToken, profile.getUserId());
+            }
+        return allUsersUpdated;
+    }
+    private static void checkUsersInSharingRegistryService(SharingRegistryServerHandler sharingRegistryServerHandler, List<UserProfile> missingUsers, String domainId) throws TException{
```
machristie: @aarushiibisht , in your testing was this method necessary? The sharing user should get created by the event based replication when enableUser is called.

aarushiibisht: @machristie ideally we should not require this method because enableUser method adds the user to user profile db which generates an event and the event is consumed by sharing registry server but while testing I found that because this is done asynchronously **sometimes** before the sharing registry processes the event and adds the user to USER_GROUP table, there is an attempt made to add user to the group. In this case it will fail because of foreign key constraint. To fix this issue I am making sure that USER_GROUP table has the user entry before adding it to the groups table. This has fixed the issue in this case but I am not sure if this scenario will arise when using portal if so we should think about fixing it. Please correct me if I am missing something.

machristie: Ok, thanks for the detailed response. That makes sense.

PR #236, opened 2019-10-23
Java 11 migration

isururanawaka:


airavata-services/services-security/src/test/java/org/apache/airavata/service/security/GatewayGroupsInitializerTest.java
```diff
@@ -37,12 +36,10 @@
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
-import org.junit.runner.RunWith;
 
 import java.util.ArrayList;
 import java.util.List;
 
-@RunWith(JMockit.class)
```
machristie: @isururanawaka I'm curious why this had to be removed and what should be done instead?

PR #243, opened 2020-03-01
Leave a comment to warn that it is not secure

YYTVicky:
Leave a comment to warn that it is not secure


Mahir92:
Hi! To get a better a understanding on what @YYTVicky referred to, if TrustManager's overridden methods have no body (i.e. - `checkClientTrusted` and `checkServerTrusted`) then all certificates will get verified (even the insecure ones).

What we propose is something like the following example code snippet - https://paste.ubuntu.com/p/jWtH2yTNR8/ . Please note that, this is just a sample code. You can use the similar format to use your customized logic to verify certificates.

If you have any questions, please me know.

PR #255, opened 2020-06-17
Changes to accommodate UserStorageQuota in StoragePreferences

vivekshresta:
This is the first cut as part of EPIC: https://issues.apache.org/jira/browse/AIRAVATA-3315. Any newly created UserStorageQuota entry for an existing StoragePreference has the value 0. 0 in the backend would mean the admin does not want any quotas/limitations for his users.

When this is the case, the frontend, instead of displaying the value 0, displays "N/A" letting know the client that there are no storage restrictions in the Gateway.


modules/registry/release-migration-scripts/next/DeltaScripts/appCatalog_schema_delta.sql
```diff
@@ -24,3 +24,18 @@ use app_catalog;
 alter table GROUP_COMPUTE_RESOURCE_PREFERENCE drop column IF EXISTS RESERVATION;
 alter table GROUP_COMPUTE_RESOURCE_PREFERENCE drop column IF EXISTS RESERVATION_START_TIME;
 alter table GROUP_COMPUTE_RESOURCE_PREFERENCE drop column IF EXISTS RESERVATION_END_TIME;
+
+-- AIRAVATA-3343: Add UserStorageQuota entry to StoragePreferences table
+SET @AddUserStorageQuota = (SELECT IF(
```
machristie, line 29: This is good, but I think it would be more readable to use the `ALTER TABLE <table name> ADD COLUMN IF NOT EXISTS` instead. I know that won't work for ide-integration since it relies on a MySQL container, but I plan to update that soon. But these release-migration-scripts we only run against MariaDB databases where IF NOT EXISTS is supported.

vivekshresta: Thanks for reviewing my code. I agree that the query isn't very readable, but sadly, as you've pointed out, unlike mariadb, mysql does not have the 'IF NOT EXISTS' syntax. 

Changing this query to a more readable MariaDB syntax throws an Error during docker-compose. One thing I can do is, write the new query and comment it, and once the migration is successful, I can remove the current complex query and uncomment the new one. Or the other approach I can think of is, remove the complex query and add a commented 'alter table STORAGE_PREFERENCE IF NOT EXISTS' query which can be uncommented when the migration is successfully merged. Please let me know if there's any better approach.

'But these release-migration-scripts we only run against MariaDB databases where IF NOT EXISTS is supported.' - Do you mean to say in the future, these migration-scripts will only be run against MariaDB?

machristie: So we have migration scripts for production MariaDB databases, thats what release-migration-scripts is. Then there are migration scripts in ide-integration, the 0*-migrations.sql scripts are.  So yes, in the 0*-migrations.sql script, IF NOT EXISTS won't work as you point out, but we can use IF NOT EXISTS in release-migration-scripts.

vivekshresta: Now I understand! Thank you for letting me know.

vivekshresta:
Hi @machristie ,

As discussed, I'm merging all my changes to a feature branch(AIRAVATA-3315-storage-quotas). Please let me know if you want me to change anything else.

machristie:
Looks good @vivekshresta

vivekshresta:
> Looks good @vivekshresta

Hi @machristie , 
Can you please merge this PR. Reaching out to you since I do not have the necessary permissions yet to merge the changes.

PR #257, opened 2020-07-27
Airavata 3348

vivekshresta:
Changes to accommodate EXPERIMENT_DATA_DIR_SIZE in the table Experiment. Thrift file changes will be merged in the next PR.


DImuthuUpe:
@vivekshresta Can you provide some context on what this PR is and what you have done so far?

vivekshresta:
> @vivekshresta Can you provide some context on what this PR is and what you have done so far?

Hi @DImuthuUpe , I'm currently working on setting quotas for different gateway users. The initial plan (out of many) when I was writing this code was to store the size of each experiment and update it whenever a new file is added to the directory(input and output files) and when a new experiment is being created, gateways call the createExperiment() on airavata where I was planning on adding some validations. The validation would involve retrieving the storage used by the user by iterating through all the experiments and getting the experimentDataDirSize and adding them. The total sum would be compared with the StorageQuota specified in the StoragePreference.

But now I feel that this might be redundant. The better approach that I feel is for the gateway to calculate the user directory size(Django-Portal has an API for calculating the size of a given directory) and while invoking the createExperiment(), send this data. Airavata then compares it with the UserStorageQuota specified in the StoragePreference and throws an exception if the quota is exceeded.

I'm currently working on that, so for now, I'll be closing this PR.

PR #260, opened 2020-08-21
Airavata 3348 final cut

vivekshresta:
This PR includes changes to validate the storage limit for a user. Airavata retrieves the size of a user given an experiment and validates against the UserStorageQuota in StoragePreference


vivekshresta:
Hi @machristie ,

Thanks for reviewing the code.

- From our last conversation in the dev mailing list, I assumed we agreed on Airavata worrying about the storage limit makes sense since, in the future, Airavata can choose between multiple StoragePreferences or choose a storage preference mentioned in GatewayResourceProfile or UserStoragePreference(which is about to be deprecated) when the storage preference id given by the gateway is invalid. I guess the gateway too can easily achieve these functionalities.

I did want the validation to happen internally, but the problem I faced was, during the experiment creation phase in Airavata, the experiment model does not have any data related to the StoragePreference in which the experiment is being created. Changing the createExperiment() method to accept another parameter would mean changes across all the gateways. And in my previous discussions with the team, I came to know that in the future, similar to choosing compute preferences for an experiment during the experiment creation phase, we're gonna develop a new functionality where the user gets to choose the StoragePreference in which he/she is going to create the experiment. With that thought process, I created a new API that can be invoked by any gateway, if they choose to use this feature. 
But I just verified that, by calling '_set_storage_id_and_data_dir(experiment)' before creating an experiment, I can set the storageId and experiment data directory removing the need for passing the storageId explicitly.

Basically, the public API can be changed to an internal API now. Will make those changes soon.


- I did consider this. The problems with this approach are: 
   1. We will check the size limit only after the experiment creation is done.
   2. When we know the size limit is exceeded, helix needs to communicate back to APIServer for deleting the created 
   experiment entries and if needed, deleting the experiment directory.
Considering these and after discussing with Dimuthu, I thought this might be the better approach when we use 'StorageResourceAdaptor', but this does seem to complicate things.

Even if I remove the new public API and integrate it with createExperiment(), this approach would still be consuming APIServer's resources(though we're using pooled resources instead of creating a new SSH connection every time). Does it make sense to just stick with the original approach - the gateway worrying about the storage quotas?
Also can you please elaborate a little on transient network failure in helix.

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            if (groupResourceProfileId == null) {
+                logger.error("Experiment not configured with a Group Resource Profile: {}", experiment.getExperimentId());
+                AiravataSystemException exception = new AiravataSystemException(AiravataErrorType.INTERNAL_ERROR);
+                exception.setMessage("Experiment not configured with a Group Resource Profile: " + experiment.getExperimentId());
+                throw exception;
+            }
+
+            GatewayResourceProfile gatewayResourceProfile = regClient.getGatewayResourceProfile(gatewayId);
+            String token;
+            if(isValid(storagePreference.getResourceSpecificCredentialStoreToken()))
```
DImuthuUpe, line 718: if (

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+                token = storagePreference.getResourceSpecificCredentialStoreToken();
+            else
+                token = gatewayResourceProfile.getCredentialStoreToken();
+
```
DImuthuUpe, line 722: Null check for the token

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
     }
 
+    @Override
+    public void validateStorageLimit(AuthzToken authzToken, ExperimentModel experiment, String storageResourceId)
+            throws InvalidRequestException, AiravataClientException, AiravataSystemException, TException {
+        String gatewayId = authzToken.getClaimsMap().get(Constants.GATEWAY_ID);
+        RegistryService.Client regClient = registryClientPool.getResource();
+        try {
+            StoragePreference storagePreference = regClient.getGatewayStoragePreference(gatewayId, storageResourceId);
+            if(storagePreference.getUserStorageQuota() == 0)
```
DImuthuUpe, line 705: if (

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+                return; //quota isn't configured
+
+            String groupResourceProfileId = experiment.getUserConfigurationData().getGroupResourceProfileId();
+            if (groupResourceProfileId == null) {
+                logger.error("Experiment not configured with a Group Resource Profile: {}", experiment.getExperimentId());
```
DImuthuUpe, line 710: Experiment {} not configured with a Group Resource Profile

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
         }
     }
 
+    @Override
+    public void validateStorageLimit(AuthzToken authzToken, ExperimentModel experiment, String storageResourceId)
+            throws InvalidRequestException, AiravataClientException, AiravataSystemException, TException {
+        String gatewayId = authzToken.getClaimsMap().get(Constants.GATEWAY_ID);
+        RegistryService.Client regClient = registryClientPool.getResource();
```
DImuthuUpe, line 702: Return the regClient back to the pool after using

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+                    storageResourceId,
+                    DataMovementProtocol.SCP,
+                    token,
+                    storagePreference.getLoginUserName());
+            String experimentDataDir = experiment.getUserConfigurationData().getExperimentDataDir();
+            String userDirectory = getUserDirectory(experimentDataDir);
+
+            //shouldn't retrieve size in giga bytes since the du -s command rounds off the value to a higher bound.
+            CommandOutput output = adaptor.executeCommand("du -sm", userDirectory);
+            if(!output.getStdError().isEmpty()) {
```
DImuthuUpe, line 734: Are you sure that output.getStdError() is not null?

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+                logger.error("The experiment data directory + " + experimentDataDir + " configured on the experiment is wrong.");
```
DImuthuUpe, line 735: Be consistent with logging. logger.error("The experiment data directory {} configured on the experiment is wrong.", experimentDataDir);

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+                    gatewayId,
+                    storageResourceId,
+                    DataMovementProtocol.SCP,
+                    token,
+                    storagePreference.getLoginUserName());
+            String experimentDataDir = experiment.getUserConfigurationData().getExperimentDataDir();
+            String userDirectory = getUserDirectory(experimentDataDir);
+
+            //shouldn't retrieve size in giga bytes since the du -s command rounds off the value to a higher bound.
+            CommandOutput output = adaptor.executeCommand("du -sm", userDirectory);
```
DImuthuUpe, line 733: Wrap this with try catch block and print as much as possible information in the catch in case of an exception. Letting this to catch at main try catch looses valuable information when debugging

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
         }
     }
 
+    @Override
+    public void validateStorageLimit(AuthzToken authzToken, ExperimentModel experiment, String storageResourceId)
```
DImuthuUpe, line 699: This method is a very risky one. You can see exceptions at most of the lines. You can use fine grained exception handling by wrapping each sub section with try catches to provide more debug information

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            CommandOutput output = adaptor.executeCommand("du -sm", userDirectory);
+            if(!output.getStdError().isEmpty()) {
+                logger.error("The experiment data directory + " + experimentDataDir + " configured on the experiment is wrong.");
+                AiravataSystemException exception = new AiravataSystemException(AiravataErrorType.INTERNAL_ERROR);
+                exception.setMessage("The experiment data directory + " + experimentDataDir + " configured on the experiment is wrong.");
+                throw exception;
+            }
+
+            double userDirectorySize = Double.parseDouble(output.getStdOut().substring(0, output.getStdOut().length() - 2).trim())/1024.0;
+            if(userDirectorySize >= storagePreference.getUserStorageQuota()) {
```
DImuthuUpe, line 742: if (

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            //shouldn't retrieve size in giga bytes since the du -s command rounds off the value to a higher bound.
+            CommandOutput output = adaptor.executeCommand("du -sm", userDirectory);
+            if(!output.getStdError().isEmpty()) {
+                logger.error("The experiment data directory + " + experimentDataDir + " configured on the experiment is wrong.");
+                AiravataSystemException exception = new AiravataSystemException(AiravataErrorType.INTERNAL_ERROR);
+                exception.setMessage("The experiment data directory + " + experimentDataDir + " configured on the experiment is wrong.");
+                throw exception;
+            }
+
+            double userDirectorySize = Double.parseDouble(output.getStdOut().substring(0, output.getStdOut().length() - 2).trim())/1024.0;
```
DImuthuUpe, line 741: Print output.getStdOut() as a log before doing this. You don't know what's happening here in case of and error

DImuthuUpe: And do you think that the output format is same for all operating systems (Linux)? If not, your algorithm might be wrong

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            if(userDirectorySize >= storagePreference.getUserStorageQuota()) {
+                logger.error(String.format("The user directory %s has exceeded the storage quota: %d GB", userDirectory, storagePreference.getUserStorageQuota()));
+                AiravataSystemException exception = new AiravataSystemException();
+                exception.setMessage(String.format("The user directory %s has exceeded the storage quota: %d GB", userDirectory, storagePreference.getUserStorageQuota()));
+                throw exception;
+            }
+        } catch (Exception ex) {
+            logger.error(ex.getMessage(), ex);
```
DImuthuUpe, line 749: More detailed error message

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            throw exception;
+        }
+    }
+
+    private String getUserDirectory(String experimentDataDirectory) {
+        //experiment directory is inside a project directory which is inside a user directory
+        String[] directories = experimentDataDirectory.split("/");
+        StringBuilder userDirectory = new StringBuilder();
+
+        for(int i = 0; i < directories.length - 2; i++)
```
DImuthuUpe, line 761: Possible index out of bound error here

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            if(!directories[i].isEmpty())
```
DImuthuUpe, line 762: if (

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
```diff
@@ -692,6 +695,80 @@ public String generateAndRegisterSSHKeys(AuthzToken authzToken, String descripti
...
+            }
+        } catch (Exception ex) {
+            logger.error(ex.getMessage(), ex);
+            AiravataSystemException exception = new AiravataSystemException(AiravataErrorType.INTERNAL_ERROR);
+            exception.setMessage(ex.getMessage());
+            throw exception;
+        }
+    }
+
+    private String getUserDirectory(String experimentDataDirectory) {
```
DImuthuUpe, line 756: This is a blind algorithm. At least provide your assumptions for experimentDataDirectory value as comments

airavata-api/airavata-api-server/src/main/java/org/apache/airavata/api/server/handler/AiravataServerHandler.java
vivekshresta: > And do you think that the output format is same for all operating systems (Linux)? If not, your algorithm might be wrong

Did not know about such a use case. Will test it and update it if necessary.

machristie:
@vivekshresta,

Regarding validating internally, I would add the check to launchExperiment instead of createExperiment. 


> We will check the size limit only after the experiment creation is done.

That's true, but the current approach only checks before the experiment runs and so doesn't account for experiment output files.

 >  When we know the size limit is exceeded, helix needs to communicate back to APIServer for deleting the created
 >   experiment entries and if needed, deleting the experiment directory.
 >   Considering these and after discussing with Dimuthu, I thought this might be the better approach when we use 'StorageResourceAdaptor', but this does seem to complicate things.

Well, my two cents, but I don't think the experiment needs to be deleted. We just need to set a flag in the database that the user is over quote on that storage resource and then prevent further file uploads/experiments on that storage resource.

>  Also can you please elaborate a little on transient network failure in helix.

Sure, Helix's task framework has builtin fault tolerance support, for example retrying in the case of failure: https://helix.apache.org/0.8.0-docs/tutorial_task_framework.html. By transient network failure I mean some sort of transient network failure between Airavata and the SSH host that prevents the SSH connection from establishing or completing successfully.

PR #265, opened 2020-10-21
AIRAVATA-3379: Enabling airavata-ruby-sdk

dinukadesilva:


ericfranz:
Some suggestions on preparing this for gem publishing to https://rubygems.org/.

Add these to the .gitignore and don't version the Gemfile.lock:

```
Gemfile.lock
.ruby-version
.ruby-gemset
```

In a tmp directory, if you do `bundle gem airavata` you can see the default gem scaffolding that is created. That is close to the process I used with the https://github.com/osc/ood_core gem. You could actually do that and copy some of the files over. If you do that, some things you will notice:

1. the Gemfile won't have any extra dependencies, the gemspec will have them all defined as either dev or runtime dependencies; thrift would become a runtime dependency
2. there is a lib/airavata.rb file that just has the `module Airavata` and a bunch of require statements at the top for all the code under the airavata directory. I think that you can get by with doing `require "airavata/airavata"` for requiring the lib/airavata/airavata.rb` file that is generated from your build process
3. you can leave version string in the gemspec or have a version file like the convention used by the scaffolding. I think the benefit of the latter is that when using the gem I can do `Airavata::VERSION` to get the version of the library I'm using programmatically, so that is nice.
4. That means that the gem version needs to [follow semantic versioning ](https://guides.rubygems.org/patterns/#semantic-versioning) and the version string of this gem should change every time you publish a new version. 
5. Not sure about providing a link back to the source code for a particular gem. I now realize that with ood_core i'm not actually doing that, though with ood_core the git tags are the almost the same as the version string (`OodCore::VERSION` is `0.14.0` and the github tag is `v0.14.0`. Rails doesn't use homepage but a separate source_code_uri though you could just use homepage - really up to you. But you can see how that works when a gem is a subdirectory of a repo: https://github.com/rails/rails/blob/v6.0.3.4/activesupport/activesupport.gemspec#L30. 
I don't know how you do that with the python sdk but would be curious to know your strategy for linking to the corresponding code in the repo. Would you create a special git tag for the entire airavata repo to reference when the new gem version is created? Or when pushing a new version of the gem, would the version match the version of Airavata itself?
5. You could have the Homepage be the source code directory at the master branch 

Main guides are https://guides.rubygems.org/make-your-own-gem/ and https://guides.rubygems.org/patterns/. You have to get accounts on rubygems.org and then if you are the first to push the gem you are the owner and then you can add owners of the gem alongside yourself.

After you are done setting things you, you can test your gem by cd-ing to the directory and running the command `bundle console`. That uses Bundler to "Start an interactive Ruby console session in the context of the current bundle".

A second way to test is to use it in another app. You can have that app's Gemfile specify the `gem "airavata", path: "/path/to/your/gem/dir"` to load that gem from the file system.

Let me know how I can help.

dinukadesilva:
@ericfranz 
This was published as a package to https://rubygems.org/gems/airavata

Thanks

dinukadesilva:
@isururanawaka @machristie 
This is ready to review

airavata-api/airavata-client-sdks/airavata-ruby-sdk/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+    $ gem install airavata
+
+## Usage
+
+```ruby
+require "thrift"
+require "airavata"
+
+transport = Thrift::BufferedTransport.new(Thrift::Socket.new('localhost', 9930))
```
isururanawaka, line 29: Is it possible to make sever configs to be loaded from a config file and hide transport creation from the user ?

dinukadesilva: Done

dinukadesilva:
As discussed, I added the utilities for authentication and transport. Also, I'm adding test cases. 

I'm having an issue with ruby thrift library. Working on troubleshooting.
```
Failure/Error: send_message('isUserExists', IsUserExists_args, :authzToken => authzToken, :gatewayId => gatewayId, :userName => userName)
     
     NoMethodError:
       undefined method `encode' for :userName:Symbol
       Did you mean?  encoding
     # ./vendor/bundle/ruby/2.3.0/gems/thrift-0.13.0/lib/thrift/bytes.rb:81:in `convert_to_utf8_byte_buffer'
     # ./vendor/bundle/ruby/2.3.0/gems/thrift-0.13.0/lib/thrift/protocol/binary_protocol.rb:110:in `write_string'
     # ./vendor/bundle/ruby/2.3.0/gems/thrift-0.13.0/lib/thrift/client.rb:44:in `write'
     # ./vendor/bundle/ruby/2.3.0/gems/thrift-0.13.0/lib/thrift/client.rb:44:in `send_message_args'
     # ./vendor/bundle/ruby/2.3.0/gems/thrift-0.13.0/lib/thrift/client.rb:30:in `send_message'
     # ./lib/airavata.rb:21:in `send_isUserExists'
     # ./lib/airavata.rb:16:in `isUserExists'
     # ./spec/sdk_spec.rb:15:in `block (2 levels) in <top (required)>'
```

dinukadesilva:
@machristie 
I tried changing the thrift version to `0.10.0` but still the same issue exists.

machristie:
Hi @dinukadesilva. As a sanity check, can you confirm that you are getting a valid access token? Is there any chance of turning on debugging or maybe changing the generated code to make sure the arguments are getting passed as expected?

The error message is rather odd since it indicates that the value for the userName is getting set to the `:userName` symbol itself.

yasithdev:
Closing this PR as it is no longer aligned with the current Airavata roadmap

PR #268, opened 2020-11-27
Fixing AIRAVATA-3391 : Resolving parameters in the groovy map data

DImuthuUpe:
The current implementation only resolves the parameters in the original groovy template. But if we pass parameters inside the map which contains parameters themselves, those are not resolved. This fix does another pass of parameter parsing on the already generated template output

Related issue: https://issues.apache.org/jira/browse/AIRAVATA-3391


machristie (approved):
Looks good to me. So the change here is to make two passes at resolving groovy parameters because some groovy map values may reference other parameters?

DImuthuUpe:
> Looks good to me. So the change here is to make two passes at resolving groovy parameters because some groovy map values may reference other parameters?

Yes that's right. One situation that might cause a trouble is referring to environment variables in the pre and post job commands. They are also prefixed using $ and I need to test what will happen in that case.

DImuthuUpe:
For environment variables, use the format \\$parameter to escape $

DImuthuUpe:
Merging as there is no outstanding concerns

PR #276, opened 2022-01-03
Fetching intermediate outputs for an running experiment

DImuthuUpe:
This PR will enable the feature of fetching output files form an already running experiment into the gateway storage

Sample client code

```
List<String> outputNames = new ArrayList<>();
outputNames.add("Gaussian-Application-Output");
apiClient.fetchIntermediateOutputs(token,"exp-id", 
        outputNames, 
        "PROCESS_98a8c45c-998c-4998-91e4-18ce915ed13e");
```


machristie (commented):
This looks really good @DImuthuUpe , thanks for getting this working.  The one comment I have is, can we derive the parentProcessId in the API or Orchestrator? It seems like we can and if so that would simplify the API for clients.  Also I'd be willing to help implement that if you want.

DImuthuUpe:
@machristie I agree with your suggestion. Possibly we can look at all the processes for the experiment and filter out the oldest one by looking at the timestamp. Please feel free to update the PR

machristie:
I've updated the PR @DImuthuUpe . 

However, I have another question. Would it possible for the process output object's value to be updated with a data product URI, similar to what is done with experiment output objects in the normal case?  I ask because otherwise clients will need to do a bit of guessing to figure out the location of the output file, especially when dealing with wildcards.

So the process output for the testing you did look like this:
```json
{
                    "name": "Gaussian-Application-Output",
                    "value": "*.log",
                    "type": 4,
                    "applicationArgument": "",
                    "isRequired": false,
                    "requiredToAddedToCommandLine": false,
                    "dataMovement": true,
                    "location": null,
                    "searchQuery": null,
                    "outputStreaming": false,
                    "storageResourceId": null,
                    "metaData": null
}
```

Where instead could the "value" be a "airavata-dp://..." data product URI, where the data product has a replication with the full path to the file?

DImuthuUpe:
Thanks @machristie. Regarding the data product url as the output, I guess it is what is supposed to happen. Have a look at

https://github.com/apache/airavata/blob/d469bf2257d179a8393f7ba6ca9060aa5d5b3fd3/modules/airavata-helix/helix-spectator/src/main/java/org/apache/airavata/helix/impl/task/AiravataTask.java#L294

Are you looking at the parent process outputs? These outputs are registered to the new process.

machristie:
Yeah you're right @DImuthuUpe , it should be setting a data product URI.  It doesn't seem to be and I can't figure out why. I'm looking at your test run yesterday, experiment id `Clone_of_Clone_of_Clone_of_Clone_of_Clone_of_Gauss_5052a8ec-0b1f-4829-ae33-af81f35e1d32`. It has two processes the job submission one `PROCESS_59024a0d-4066-41b5-953a-caa2c8a59fc9` and the intermediate output fetching one `PROCESS_7ff674cb-d9ba-4805-be06-4bf6f5dfbef6`. In the latter, there is just the one process output, for `Gaussian-Application-Output` and it's value is `*.log`.  

```
...
"processOutputs": [
                {
                    "name": "Gaussian-Application-Output",
                    "value": "*.log",
                    "type": 4,
                    "applicationArgument": "",
                    "isRequired": false,
                    "requiredToAddedToCommandLine": false,
                    "dataMovement": true,
                    "location": null,
                    "searchQuery": null,
                    "outputStreaming": false,
                    "storageResourceId": null,
                    "metaData": null
                }
            ],
...
```
For reference, I'm seeing this through the django.seagrid.org REST API: https://django.seagrid.org/api/experiments/Clone_of_Clone_of_Clone_of_Clone_of_Clone_of_Gauss_5052a8ec-0b1f-4829-ae33-af81f35e1d32/

machristie:
@DImuthuUpe  I think I've figure it out. Gaussian-Application-Output has type URI_COLLECTION (which might be a misconfiguration, but we'll leave that aside for now). So when the output objects are saved, AiravataTask is calling saveExperimentOutputCollection https://github.com/apache/airavata/blob/AIRAVATA-3549/modules/airavata-helix/helix-spectator/src/main/java/org/apache/airavata/helix/impl/task/AiravataTask.java#L306-L329. I think this method also needs to be updated like the way saveExperimentOutput was updated to save outputs to either the experiment or process.

DImuthuUpe:
@machristie good catch. I will update the code. Thanks  for this

machristie:
Hi @DImuthuUpe , I have one other request. I noticed that the output fetching processes have ProcessStatuses of STARTED (2) and then OUTPUT_DATA_STAGING (8). It would be nice if they also had a final ProcessStatus of COMPLETED or FAILED.

I can work around this by looking at TaskStatus, but it would be nicer to have a final COMPLETED/FAILED status at the Process level.

machristie:
I've added some validations to the API method
- user has WRITE access to experiment
- experiment is EXECUTING
- there are no currently executing intermediate output fetching processes for the given outputNames

I also added a new API method `getIntermediateOutputProcessStatus(AuthzToken authzToken, String airavataExperimentId, List<String> outputNames)` which returns a ProcessStatus object and indicates the status of a fetching intermediate outputs. This will make it easier on clients to figure out when an intermediate output fetching process is done.

machristie:
@DImuthuUpe FYI, I'm going to work on updating saveExperimentOutputCollection to update the Process outputs. I'll also see if I can add a final COMPLETED/FAILED status to the output fetching processes.

machristie:
Changed the logic for allowing fetch of intermediate outputs to require that the job itself is ACTIVE. Previously it was only checking that the Experiment was in EXECUTING state. This was a good suggestion from @eroma2014.

PR #278, opened 2022-01-14
Relevant changes for Condor job submissions

DImuthuUpe:
https://github.com/apache/airavata/tree/htcondor branch contains the working code for condor submissions but it also contains many unrelated changes that need to be addressed separately. So this one is just a cut and paste of the condor related changes from that branch


dev-tools/ansible/roles/job_monitor/templates/email-monitor/email-config.yaml.j2
```diff
@@ -93,3 +93,7 @@ config:
      resourceEmailAddresses:
        - iu.xsede.edu # test resource mail address
        - tcs.tulsahpc.org #Tandy
+
+   - jobManagerType: HTCondor
```
dinukadesilva, line 97: This jobManagerType has to be equal to the enum since it's already referenced at https://github.com/apache/airavata/blob/a37ae49ef4959c731dd220830711bd790c7e1635/modules/airavata-helix/helix-spectator/src/main/java/org/apache/airavata/helix/impl/task/submission/config/GroovyMapBuilder.java#L80

DImuthuUpe: No. For email monitor, parser is fetch from this line https://github.com/apache/airavata/blob/htcondor-new/modules/job-monitor/email-monitor/src/main/java/org/apache/airavata/monitor/email/EmailBasedMonitor.java#L130. So the value in email-config.yml should be "HTCondor" as it is pointing to the parser https://github.com/apache/airavata/blob/htcondor-new/modules/job-monitor/email-monitor/src/main/java/org/apache/airavata/monitor/email/parser/HTCondorEmailParser.java

However, the HTCondo custom override in GrrovyMapBuilder should be removed as it kills the generic behavior of the code.

PR #279, opened 2022-02-02
Enabling HTCONDOR job submission option on pga scigap portal

dinukadesilva:


DImuthuUpe:
@dinukadesilva You have to send the PR to https://github.com/apache/airavata-php-gateway/tree/develop. And this file is a thrift generated file. You might have to build it using https://github.com/apache/airavata/blob/htcondor-new/thrift-interface-descriptions/generate-thrift-stubs.sh and copy those Airavata php thrift stubs to php-gateway code base

PR #281, opened 2022-03-17
AIRAVATA-3590 upgrade log4j and commons-io

pjfanning:


smarru:
Some of these may be obsolete, but will merge these and will later review and remove any unused dependencies.

PR #282, opened 2022-03-17
AIRAVATA-3590: jackson 2.9.10.8

pjfanning:


smarru:
Thank you @pjfanning for your pro-active review, I see no immediate concerns, so will merge these and will test through if the dependency upgrades causes any issues.

PR #288, opened 2022-04-05
AIRAVATA-3592 use jackson2

pjfanning:


pjfanning:
@smarru could you review this if you time?

PR #289, opened 2022-04-05
AIRAVATA-3590 remove xmlbeans

pjfanning:
XMLBeans 2.5.0 has a serious XXE issue and does not really seem to be used anyway


smarru:
Thanks @pjfanning yes this is one my chopping block, will merge this

PR #298, opened 2022-06-09
Handling special characters in output staging

DImuthuUpe:
In some cases output files might have special characters in the name which causes trouble for SCP protocol to handle. This fix will rename those files


machristie (commented):
@DImuthuUpe This looks good to me. One question I have, do you need to escape special characters in passThroughTransfer too? Or does this only affect naiveTransfer?

DImuthuUpe:
@machristie thanks for the review. passThroughTransfer is obsolete right now. This is should be replaced with MFT integration in future.

machristie:
@DImuthuUpe sounds good, thanks for the explanation

PR #329, opened 2022-10-17
Airavata Metascheduler Project Structure

isururanawaka:


machristie:
@isururanawaka looks like you're still working on this? Let me know when you want me to review it.

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/entities/expcatalog/ComputationalResourceSchedulingEntity.java
```diff
@@ -0,0 +1,162 @@
...
+import javax.persistence.*;
+import java.io.Serializable;
+
+/**
+ * Persistent class for computational_resource_scheduling data table.
+ */
+@Entity
+@Table(name = "COMPUTE_RESOURCE_SCHEDULING")
+@IdClass(ComputationalResourceSchedulingPK.class)
+public class ComputationalResourceSchedulingEntity implements Serializable {
```
machristie, line 12: I don't think this is needed. There's a mismatch between the Thrift data model and the database entities here. UserConfigurationDataModel and ComputationalResourceSchedulingModel map to the single UserConfigurationDataEntity/table in the database.

modules/airavata-metascheduler/process-scheduler/src/main/java/org/apache/airavata/metascheduler/process/scheduling/engine/cr/selection/ComputeResourceSelectionPolicyImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        }finally {
+            this.registryClientPool.returnResource(client);
+        }
+    }
+
+    public String getComputeResourceCredentialToken(String gatewayId,
+                                                    String username, String computeResourceId, boolean isUseUserCRPref,
+                                                    boolean isSetGroupResourceProfileId, String groupResourceProfileId) throws Exception {
+        if (isUseUserCRPref) {
+            if (getUserComputeResourcePreference(gatewayId, username, computeResourceId) != null &&
```
machristie, line 52: getUserComputeResourcePreference will never return null because RegistryService.Client.getUserComputeResourcePreference() will never return null. Thrift RPC methods can't return null and instead will throw an exception.  In getUserComputeResourcePreference() I think you could catch the exception and return null, but ideally we'd have a Thrift method that does the existence check and returns a boolean (like RegistryServerHandler.isUserResourceProfileExists).

machristie: likewise for 
- getUserResourceProfile
- getGroupComputeResourcePreference
- getGroupResourceProfile

airavata-api/airavata-data-models/src/main/java/org/apache/airavata/model/status/ProcessState.java
```diff
@@ -42,7 +42,10 @@ public enum ProcessState implements org.apache.thrift.TEnum {
   COMPLETED(10),
   FAILED(11),
   CANCELLING(12),
-  CANCELED(13);
+  CANCELED(13),
```
machristie, line 45: You must have updated status_models.thrift, but that file is missing from the PR.

machristie (changes requested):
Hi @isururanawaka , this is looking like a pretty good structure. I left some comments on the PR.

I guess for right now this implementation checks that compute resources are actually up and available. In the future will it also automatically select a compute resource on which to run an application?

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/entities/expcatalog/ComputationalResourceSchedulingEntity.java
machristie: @isururanawaka no I'm saying this Entity class isn't needed. These fields from the Thrift model are saved into the user_configuration_data table.

PR #361, opened 2022-12-27
multiple job stacks can be created for autoscheduled enabled processes

isururanawaka:


isururanawaka:
@DImuthuUpe can this create side effects for general job submissions. It allows creating multiple job stacks for same processes if autoscheduling is enabled in case of failed processes at job submission phase.This keeps resubmitting jobs to cluster with new jobid for requeued processes.

DImuthuUpe:
@isururanawaka This check is performed because there was a bug in helix core to repeatedly run the same task without termination. In such cases, there were lots of jobs with the same configuration created for the same task. If you put this logic here, there is a possibility of that bug causing the same issue for auto-scheduled processes. I do not have a better way to write this but consider the issue I mentioned and you should be able to write this in a different way while mitigating the issue.

isururanawaka:
thanks @DImuthuUpe .

Do you know whether helix cores retries with the same task object repeatedly or create a new task object? BTW unsuccessful job stacks are deleted for now except the latest one.

PR #381, opened 2023-02-02
add time distribution calculation

isururanawaka:


DImuthuUpe:
@isururanawaka I understand that these utilities are required for the performance measurement of your testing. I am still wondering what is the use case of exposing those through the registry

isururanawaka:
> @isururanawaka I understand that these utilities are required for the performance measurement of your testing. I am still wondering what is the use case of exposing those through the registry

@DImuthuUpe I'm thinking of some stat modules will be helpful for meta scheduler to design routing strategies in the future(https://github.com/apache/airavata/tree/develop/modules/airavata-metascheduler/metadata-analyzer). But, for now, I'm using this only for testing. once, done we can disable it.

PR #433, opened 2023-08-12
Airavata 3594 admin dashboard additions

SauravKumarJhaNITW:
added a method to return UserGroups filtered by creation date


modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/repositories/expcatalog/JobStatusRepository.java
```diff
@@ -113,4 +117,51 @@ public List<JobStatus> getDistinctListofJobStatus(String gatewayId,String status
...
+        List<CpuUsage> cpuUsages = new ArrayList<>();
+        List<JobStatus> startedJobStatusList = jobStatusRepository.selectWithNativeQuery(
+                QueryConstants.FIND_JOB_STATUS_STARTED_BEFORE_TIME, gatewayId, JobState.ACTIVE.name(),
+                String.valueOf(toTime));
+        ;
+        List<JobStatus> finishedJobStatusList = jobStatusRepository.selectWithNativeQuery(
+                QueryConstants.FIND_JOB_STATUS_FINISHED_AFTER_TIME, gatewayId, JobState.COMPLETE.name(),
+                JobState.FAILED.name(), JobState.CANCELED.name(), String.valueOf(fromTime));
+        Mapper mapper = ObjectMapperSingleton.getInstance();
+        for (JobStatus startedJobStatus : startedJobStatusList) {
```
machristie, line 132: This double for loop isn't going to scale well for gateways with a large number of jobs. It would be better to do this all in the query. I would suggest something like this:

```sql
select j.* from JOB j
inner join PROCESS p on p.PROCESS_ID = j.PROCESS_ID
inner join EXPERIMENT e on e.EXPERIMENT_ID = p.EXPERIMENT_ID
where e.GATEWAY_ID = ?
and exists (select 1 from JOB_STATUS js where js.JOB_ID = j.JOB_ID and js.STATE = ? and js.TIME_OF_STATE_CHANGE <= ?)
and exists (select 1 from JOB_STATUS js where js.JOB_ID = j.JOB_ID and (js.STATE = ? or js.STATE = ? or js.STATE = ?) and js.TIME_OF_STATE_CHANGE >= ?)
```

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/repositories/expcatalog/JobStatusRepository.java
```diff
@@ -113,4 +117,51 @@ public List<JobStatus> getDistinctListofJobStatus(String gatewayId,String status
...
+            for (JobStatus finishedJobStatus : finishedJobStatusList) {
+                JobStatusEntity startedJobStatusEntity = mapper.map(startedJobStatus, JobStatusEntity.class);
+                JobStatusEntity finishedJobStatusEntity = mapper.map(finishedJobStatus, JobStatusEntity.class);
+                if (startedJobStatusEntity.getJobId() == finishedJobStatusEntity.getJobId()) {
```
machristie, line 136: I'm pretty sure this won't work. Mapping a JobStatus back to an Entity won't populate the extra fields (like `getJobId()`) that are stored on the database. But see my comment above, if that approach works you won't need to do this.

modules/registry/registry-core/src/main/java/org/apache/airavata/registry/core/repositories/expcatalog/JobStatusRepository.java
```diff
@@ -113,4 +117,51 @@ public List<JobStatus> getDistinctListofJobStatus(String gatewayId,String status
...
+                    String jobId = startedJobStatusEntity.getJobId();
+                    List<ExperimentModel> experiments = (new ExperimentRepository())
+                            .selectWithNativeQuery(QueryConstants.FIND_EXPERIMENT_WITH_JOB_ID, jobId);
+                    if (!experiments.isEmpty()) {
+                        ExperimentModel experiment = experiments.get(0);
+                        List<Integer> res = (new ExperimentRepository().selectWithNativeQuery(
```
machristie, line 142: Do you need this query? You've already loaded the ExperimentModel so I think you can just use `experiment.getUserConfigurationData().getComputationalResourceScheduling().getTotalCPUCount()`.

machristie (changes requested):
Hi @SauravKumarJhaNITW , thank you for this contribution! Overall this looks really good. I do have some requested changes in my review mainly around making sure that the getCpuUsage code can scale to handle gateways with thousands of jobs. Please let me know if you have any questions.

PR #453, opened 2025-01-22
Fix: AIRAVATA-3001 - Remove multiple INSTALL files and NOTICE files

ichandrasharma:
Redundant INSTALL and NOTICE files in source and binary distributions lead to out-of-sync information. Merge both of them into a single location and clearly distinguish the source and binary information and NOTICE files

this pull now resolves the issue by consolidating both the installation and notice details into a single location


Blevins42069:
export LD_LIBRARY_PATH=/usr/local/cuda/lib64

PR #456, opened 2025-02-20
Updated keycloak image in docker-compose.yml

katamyra:
Updated Keycloak container image from `jboss/keycloak:2.5.4.Final` to `quay.io/keycloak/keycloak:latest`. I believe the official Keycloak container image was moved away from jboss/keycloak, and it was causing errors when running `docker-compose up` in the ide-integration guide.

### DIFF
--- a/modules/ide-integration/src/main/containers/docker-compose.yml
+++ b/modules/ide-integration/src/main/containers/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   keycloak:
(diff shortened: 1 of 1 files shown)


yasithdev:
Hi @katamyra,

Thanks for the commit! The keycloak version in the root `pom.xml` is `24.0.4`. Could you use this version instead? Also, new issues might arise after you bump the keycloak version, so can you confirm if the services run as expected after this upgrade?

PR #457, opened 2025-02-23
Updated Python SDK

yasithdev:
This PR contains an updated Python SDK, with the following improvements.
- regenerated code using Thrift 0.21.0, which adds enums and type hints.
- Updated __init__.py files to import missing modules ONLY when type-checking (otherwise circular imports occur). This step prevented VSCode and PyCharm from complaining about broken imports within the SDK.
- Removed redundant wrapper code around the thrift-generated SDK. This greatly improved IDE stability and performance during static type checking.


lahirujayathilake:
@yasithdev LGTM 👍

PR #462, opened 2025-03-11
AIRAVATA-3962: Improved docker-compose.yml for stability and compatibility

Bcoderx6:
The current docker-compose.yml file in Apache Airavata has several issues affecting stability and compatibility:

Keycloak uses an outdated startup command (-b 0.0.0.0), which is no longer supported in newer versions.
RabbitMQ does not have the management plugin enabled, making it inaccessible on port 15672.
No clear dependency order for some services.

 ✅ Proposed Fix
This issue proposes updating docker-compose.yml to:
Use RabbitMQ 3.12 with management plugin for better UI support.
Fix Keycloak startup command for newer versions.
Ensure correct startup order using depends_on.

📌 Testing Done
Successfully tested all services using curl and nc commands.
Verified Keycloak UI at http://localhost:18080/.
Verified RabbitMQ UI at http://localhost:15672/. 

image
@yasithdev

### DIFF
--- a/modules/ide-integration/src/main/containers/docker-compose.yml
+++ b/modules/ide-integration/src/main/containers/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   keycloak:
-    image: quay.io/keycloak/keycloak:latest
+    image: quay.io/keycloak/keycloak:24.0.0
     environment:
       - KEYCLOAK_USER=admin
       - KEYCLOAK_PASSWORD=admin
--- a/modules/ide-integration/src/main/containers/docker-compose.yml
+++ b/modules/ide-integration/src/main/containers/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   keycloak:
-    image: jboss/keycloak:2.5.4.Final
+    image: quay.io/keycloak/keycloak:latest
     environment:
       - KEYCLOAK_USER=admin
       - KEYCLOAK_PASSWORD=admin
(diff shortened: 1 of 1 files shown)


lahirujayathilake:
@Bcoderx6, Thanks for initiating this. As Dimuthu mentioned, you must follow the entire ide-integration to verify the upgrades are compatible with the Airavata services. (including Airavata Django portal)
https://github.com/apache/airavata/blob/master/modules/ide-integration/README.md?plain=1#L68

I can confirm the Airavata services were working with the Keycloak `24.0.0` version

Bcoderx6:
@lahirujayathilake , thanks for your feedback!  

I see that Keycloak 24.0.0 is a known stable version for Airavata. Currently, the latest Keycloak version is working without any issues in my tests.  

However, I will update it to 24.0.0 because using the latest tag in Docker pulls new versions automatically whenever Keycloak releases an update. This could introduce unexpected issues. Instead, it's better to test new Keycloak versions first and then update docker-compose.yml accordingly.  

Let me know if you have any concerns! , Thank You

piyushupadhyay19:
@Bcoderx6 Thanks for updating jboss/keycloak to the Quarkus based keycloak

However, I think the old variables (KEYCLOAK_USER and KEYCLOAK_PASSWORD) no longer work in here. Instead, we should use:

KEYCLOAK_ADMIN
KEYCLOAK_ADMIN_PASSWORD

Without this change, admin login fails at http://localhost:18080/. I ran into this issue while testing and updating the env variables fixed it.
@lahirujayathilake

lahirujayathilake:
@Bcoderx6 and @piyushupadhyay19, thanks for the updates. And FYI - https://github.com/apache/airavata/pull/464

Bcoderx6:
@lahirujayathilake Sure Lahiru, sorry for the late reply. I just finished my final exam today. I'll also go ahead and close this PR.

PR #466, opened 2025-03-22
use fasterxml import instead of codehaus

blitzboah:
issue link - https://issues.apache.org/jira/browse/AIRAVATA-3592

- replaced codehaus import with fasterxml

didn't changed pom.xml as it already has fasterxml


lahirujayathilake:
@blitzboah, this change would cause runtime issues because fasterxml is required by the current version of helix. 
There are some active developments happening on a different branch to handle version upgrades - https://github.com/apache/airavata/blob/cybershuttle-dev/pom.xml#L151

PR #490, opened 2025-04-11
Support session termination & deletion on jupyter

ganning127:


DImuthuUpe (approved):
Thanks @ganning127. This is looking great

PR #513, opened 2025-06-08
Remove Static Thrift Stubs and Set Up Auto Generation at Build TIme

yasithdev:
This PR removes several auto-generated thrift stubs (java) from the codebase, and auto-generates them (in target/generated-sources) at build time. It also makes the stubs and java code consistent with each other.

* Generates Thrift code at compile time instead of storing pre-generated snapshots
* Reduces codebase size and removes duplication
* Ensures consistency between Thrift files, generated code, and SDKs
* Makes it easier to track API versions across languages


lahirujayathilake:
@yasithdev, LGTM 👍

PR #522, opened 2025-06-17
Support resource deletion

ganning127:
Addresses: https://github.com/cyber-shuttle/cybershuttle/issues/13

### Behavior
When a resource / project is "deleted", we simply set the "state" property on that hibernate entity to be DELETED. If we were to delete the hibernate entity completely, that would mess up foreign key relationships with already existing projects that depend on a deleted resource. Or, completely deleting a project entity may create issues with sessions, which may be created by other users. Thus, we simply mark the project/resource as deleted, and never return it in any future search results

### Resource Deletion
https://github.com/user-attachments/assets/624e4b33-0c4f-4e3d-bf71-f69d63e8557f

### Project Deletion
https://github.com/user-attachments/assets/da3b646c-58bc-47ef-8967-8142b19c32fd

### DIFF
--- a/modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/controller/ResourceController.java
+++ b/modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/controller/ResourceController.java
@@ -1,53 +1,41 @@
 /**
-*
-* Licensed to the Apache Software Foundation (ASF) under one
-* or more contributor license agreements. See the NOTICE file
-* distributed with this work for additional information
-* regarding copyright ownership. The ASF licenses this file
-* to you under the Apache License, Version 2.0 (the
-* "License"); you may not use this file except in compliance
-* with the License. You may obtain a copy of the License at
-*
-* http://www.apache.org/licenses/LICENSE-2.0
-*
-* Unless required by applicable law or agreed to in writing,
-* software distributed under the License is distributed on an
-* "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-* KIND, either express or implied. See the License for the
-* specific language governing permissions and limitations
-* under the License.
-*/
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership. The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
 package org.apache.airavata.research.service.controller;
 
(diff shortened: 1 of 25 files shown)


modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/controller/ResourceController.java
```diff
@@ -29,25 +29,13 @@
 import org.apache.airavata.research.service.enums.ResourceTypeEnum;
 import org.apache.airavata.research.service.handlers.ProjectHandler;
 import org.apache.airavata.research.service.handlers.ResourceHandler;
-import org.apache.airavata.research.service.model.entity.DatasetResource;
-import org.apache.airavata.research.service.model.entity.ModelResource;
-import org.apache.airavata.research.service.model.entity.NotebookResource;
-import org.apache.airavata.research.service.model.entity.Project;
-import org.apache.airavata.research.service.model.entity.RepositoryResource;
-import org.apache.airavata.research.service.model.entity.Resource;
+import org.apache.airavata.research.service.model.entity.*;
```
lahirujayathilake, line 32: @ganning127 revert these changes. Don't use wild card imports

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/handlers/ResourceHandler.java
```diff
@@ -177,6 +180,23 @@ public Resource getResourceById(String id) {
         return opResource.get();
     }
 
+    public boolean deleteResourceById(String id) {
+        Optional<Resource> opResource = resourceRepository.findById(id);
+
+        if (opResource.isEmpty()) {
+            throw new RuntimeException("Resource not found: " + id);
```
lahirujayathilake, line 187: use `jakarta.persistence.EntityNotFoundException` and better to log this

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/handlers/ResourceHandler.java
```diff
@@ -177,6 +180,23 @@ public Resource getResourceById(String id) {
...
+        }
+
+        String userEmail = UserContext.userId();
+        System.out.println("Received request to delete with user: " + userEmail);
```
lahirujayathilake, line 191: remove system outs, if needed either use info or debug log

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/handlers/ResourceHandler.java
```diff
@@ -177,6 +180,23 @@ public Resource getResourceById(String id) {
...
+        if (!opResource.get().getAuthors().contains(userEmail.toLowerCase())) {
```
lahirujayathilake, line 192: @ganning127 shall we add a TODO comment here saying that only the owner should be able to delete a resource?
Once we integrate roles and sharing we should restrict this feature

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/handlers/ResourceHandler.java
```diff
@@ -177,6 +180,23 @@ public Resource getResourceById(String id) {
...
+            throw new RuntimeException("User " + userEmail + " not authorized to delete resource");
```
lahirujayathilake, line 193: better to log these, indicating resource type, id, and user

lahirujayathilake (changes requested):
@ganning127, I added a couple of minor comments. Once they are resolved we can merge this

lahirujayathilake:
Thanks @ganning127 LGTM 👍

PR #527, opened 2025-06-20
Support liking resources

ganning127:
Added:
- ability to like all resources
- page to view liked resources

https://github.com/user-attachments/assets/d97edb68-e8e8-4429-98ba-8c53e0341d43

Addresses: https://github.com/cyber-shuttle/cybershuttle/issues/18


modules/research-framework/portal/src/components/resources/LikeResourceButton.tsx
```diff
@@ -0,0 +1,84 @@
+import {Resource} from "@/interfaces/ResourceType.ts";
```
lahirujayathilake, line 1: include the license header

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/controller/LikesController.java
```diff
@@ -0,0 +1,66 @@
+/**
```
lahirujayathilake, line 1: I believe it is better to move all the endpoints to the existing resource controller instead of creating a separate controller class

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/controller/ResourceController.java
```diff
@@ -57,7 +57,7 @@ public class ResourceController {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ResourceController.class);
 
-    @org.springframework.beans.factory.annotation.Autowired
+    @Autowired
```
lahirujayathilake, line 60: better to use constructor-based dependency injection as a practice

modules/research-framework/research-service/src/main/java/org/apache/airavata/research/service/model/entity/ResourceLike.java
```diff
@@ -0,0 +1,70 @@
...
+public class ResourceLike {
+
+    @Id
+    @GeneratedValue
+    @UuidGenerator
+    @Column(nullable = false, updatable = false, length = 48)
+    private String id;
+
+    @Column(name = "owner_id", nullable = false)
+    private String ownerId;
```
lahirujayathilake, line 41: @ganning127, this is to hold the id of the user who liked the resource right? Better to change the variable name

modules/research-framework/portal/src/components/resources/LikeResourceButton.tsx
lahirujayathilake, line 1: I prefer to change the wording to 'Star' instead of 'Like'. WDYT @yasithdev?

lahirujayathilake (changes requested):
@ganning127 I added some comments.

modules/research-framework/portal/src/components/resources/LikeResourceButton.tsx
yasithdev: Yes, agreed!

PR #528, opened 2025-06-21
Show error toast if user navigates to a resource with invalid ID

ganning127:
Currently, if a user tries to access a resource with invalid id, they will be shown an infinite loading state. This PR fixes that.

Addresses https://github.com/cyber-shuttle/cybershuttle/issues/34


lahirujayathilake:
LGTM 👍

PR #550, opened 2025-07-29
Backend support for listing tags in alphabetical order

ganning127:
Addresses: https://github.com/cyber-shuttle/cybershuttle/issues/48

Screenshot 2025-07-28 at 11 23 23 PM


airavata-api/src/main/java/org/apache/airavata/helix/impl/task/aws/AWSCompletingTask.java
```diff
@@ -1,21 +1,22 @@
 /**
- * Licensed to the Apache Software Foundation (ASF) under one
```
ganning127, line 2: The only changes to this file are from mvn spotless:apply.

Otherwise, the build won't pass

PR #561, opened 2025-10-06
Support consecutive batch job runs in agent

lahirujayathilake:


DImuthuUpe:
@lahirujayathilake let's move this to linkspan

